PR #520, opened 2026-08-20
feat: LoRA Sorter tab — sort LoRAs into base-model/category folders

Little-God1983:
## LoRA Sorter — third tab in the LoRA Viewer

Sorts the LoRA library into clean folders by base model and (optionally) category, with a live folder-structure preview, move/copy modes, disk-space pre-flight, fully automatic collision handling, and a per-run sort-history manifest.

Spec: `docs/superpowers/specs/2026-08-20-lora-sorter-design.md` · Plan: `docs/superpowers/plans/2026-08-20-lora-sorter.md`

### What's in it
- **Sort your LoRAs tab** (after Browse Civitai): options rail (source incl. Browse-any-folder, target, structure, move/copy, delete-empty-folders, Refresh) + "Folder structure preview" tree (recursive DataTemplates, no TreeView) + disk gate (required vs free, 1 GB margin) + busy overlay with working Cancel.
- **DB-driven planner** (`LoraSortPlanner`, pure): candidates from the cached library graph; category via `SorterCategoryResolver` — byte-for-byte the downloader's tag inference, so sorted files and fresh downloads land in identical folders.
- **Collision policy (no dialogs, no overwrite path anywhere):** different content → deterministic `{stem}_{versionId}` rename (downloader convention → idempotent re-runs); identical content → skipped + reported (Find Duplicates remains the dedup tool).
- **Executor:** sidecars travel with the model file (incl. renames), `overwrite: false` on every operation, `ModelFile.LocalPath` repointed in batched fresh-UoW scopes on move (copy leaves DB untouched), survives cancellation/DB failures mid-run, per-run manifest in `%LocalAppData%\DiffusionNexus\SortHistory\` (data source for a future Restore).
- **Arbitrary folders:** unknown files resolved via sidecar → per-hash cache (`%LocalAppData%\DiffusionNexus\SorterCache\`) → Civitai by-hash; 404s negatively cached; offline never throws.
- All heavy work (hashing, enumeration, execution) off the UI thread with marshalled progress; access-denied subtrees skipped, preview errors surfaced to the status bar; Unified Console logging (`FileSystem`/`LoraSorter`) throughout.

### Verification
- 73 new tests (planner/path/category/sidecar/manifest/executor/resolver/VM), full suite **3893 green**, UI Release build clean.
- Ten task-scoped reviews + adversarial final whole-branch review; all Critical/Important findings fixed (incl. UI-thread offload, cancellable resolve, stale-banner fix).

### Known follow-ups (deliberately deferred)
- DB-by-hash resolver step descoped for v1 (browsed-folder duplicates hit the API once, then cached).
- `BuildPlan`'s lazy-hash phase isn't cancellable between resolve and commit.
- `InaccessibleSubfolderIsSkippedNotFatal` is a no-throw smoke, not a real ACL test.
- Restore UI (manifest is already written every run).

### Manual smoke checklist (owed before merge)
- [ ] Move run (category structure): files + sidecars land, Installed tab paths current, manifest written
- [ ] Copy run to a separate folder: source + DB untouched
- [ ] Multi-GB copy: overlay animates, progress live, Cancel responsive (UI-thread offload witness)
- [ ] Cancel mid-run: completed files stay, partial reported
- [ ] Cancel during metadata resolution of a browsed folder
- [ ] Browsed folder with an access-denied subfolder: skipped, not fatal
- [ ] Expand a folder with 2+ files: totals correct (lazy-realization check)
- [ ] Re-run same sort: everything "already in place", zero transfers


Little-God1983:
## Code review — LoRA Sorter

Review-only pass at `high` effort. **No fixes applied**; nothing on the branch was touched.

**Verification context:** full solution built in Release (clean) and the suite run — 3893 tests, one failure (`GenerationGalleryViewModelTests.TagCloudSearchText_FiltersTheVisibleChips`), an unrelated parallel-execution flake that passes in isolation and touches no file in this PR. Two findings below were confirmed by running real `dotnet` probes rather than by reading code.

Each finding is tagged **[confirmed]** (traced end-to-end through the code, or verified empirically) or **[plausible]** (surfaced by a finder pass, not independently re-verified — worth a look before acting).

---

### Overall

The architecture is sound and matches the approved spec: the plan/execute split, the DB-repoint batching, the preview tree, the no-overwrite collision policy and the busy/cancel plumbing all hold up, and happy-path test coverage is genuinely good. The 24 findings cluster in three places, and only the first two are the kind of thing that should block a merge.

1. **Error paths leave the ViewModel armed with stale state.** Preview passes write their results only on the success path, so a failed or cancelled pass leaves `Start` enabled against the *previous* folder's plan.
2. **Exception filters are narrower than the exceptions the code can actually raise**, and the terminal DB flush is not in a `finally` — the failure mode is moved files with stale DB rows.
3. **Parity with the download pipeline is asserted in comments, not enforced in code** — and the two already disagree today, in a way that makes sorter and downloader fight over the same files forever.

---

## 1. Stale state on error paths

#### 1.1 A failed or cancelled preview leaves the previous plan armed — [plausible]

[`LoraSorterViewModel.cs:569`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L569)

`_lastPlan`, `TransferCount` and `HasEnoughSpace` are written only at lines 399/403/407, i.e. only when a pass completes. Neither the `OperationCanceledException` path (line 378 `return`) nor the general catch (line 433) resets them, and `RunBusyAsync`'s `finally` clears `IsBusy` regardless — so `CanStart` re-arms with the old values.

Concretely: preview S1 succeeds (`planA`, `IsMove: true`, 42 transfers). User switches to S2, or flips to Copy; that pass throws — a locked file during collision hashing, or a `LoadCachedFilesAsync` failure, which the PR's own `PreviewFailureIsSurfacedNotSwallowed` test proves is a reachable path. The button still reads "Start Sorting (42 files)". The confirm dialog interpolates the **live** `IsMove`/`EffectiveTargetRoot` ("42 files will be copied into S2"), the `ReferenceEquals(_lastPlan, plan)` guard passes because nothing reassigned `_lastPlan`, and the executor then runs `planA` with `plan.IsMove == true` — files are **moved** inside S1 after the user was told they would be copied into S2. Line 661's `if (IsMove && DeleteEmptySourceFolders)` then prunes S2, a folder the run never touched.

The `ReferenceEquals` guard was clearly written for this class of problem, but it cannot see this one.

#### 1.2 One undeletable folder turns a fully successful sort into "Sorting failed" — [confirmed]

[`LoraSorterViewModel.cs:663`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L663)

`DiskUtility.DeleteEmptyDirectories` calls `Directory.Delete` with no try/catch of its own, and the call site adds none. An Explorer window or AV holding one now-empty folder, or a permission-denied subdirectory (exactly the case `EnumerateFilesSafe` deliberately tolerates), throws *after* `taskHandle.Complete(...)` already reported success. It unwinds into `StartSortingAsync`'s catch at 620, so lines 606–618 never run: `_candidateCache` is not cleared, no post-run recompute happens, and `SortCompleted` never fires so the Installed tab never refreshes. Every file moved correctly, but the user sees `Sorting failed: Access to the path … is denied` and a preview still listing the old, now-nonexistent paths. Pressing Start replays a plan whose sources are gone.

---

## 2. Exception filters and abort semantics

#### 2.1 The final DB flush is not in a `finally` — [confirmed]

[`LoraSortExecutor.cs:123`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortExecutor.cs#L123)

The transfer loop's filter is `when (ex is IOException or UnauthorizedAccessException)`. `_historyWriter.MarkCompleted` (line 118) sits inside that try and does a full read-deserialize-write of the manifest per file, so it can raise a `JsonException` (truncated manifest), a `NullReferenceException` (the `!` on a null deserialize result), or an `ArgumentOutOfRangeException` from `SidecarLocator.DeriveSidecarTargetPath`'s unchecked `sidecarName[sourceStem.Length..]` slice. None match the filter; all unwind straight out of `ExecuteAsync`.

Because the `if (pendingDbChanges.Count > 0)` block sits *after* the loop rather than in a `finally`, up to 19 pending path repoints are silently dropped. Those files are physically at their new locations while the DB still points at the old ones — they show as missing in the Installed tab. A milder variant: a transient `IOException` from `MarkCompleted` *is* caught, and counts a file that actually moved as `failed`, so the report contradicts both disk and DB.

#### 2.2 The DB-batch failure handler never clears the batch — [confirmed]

[`LoraSortExecutor.cs:110`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortExecutor.cs#L110)

The `continue` skips `MarkCompleted`, `done++` and `progress.Report` for a file that *was* transferred, and `pendingDbChanges.Clear()` (line 96) is never reached — so the list stays at 20 and the very next successful transfer trips `>= DbBatchSize` and re-throws. For every remaining file.

On a 3000-file move that fails at file 20: the progress bar freezes at ~0.6% for the rest of the run, the manifest records almost nothing as completed (poisoning the Restore feature it exists for), and the DB takes ~2980 failing round-trips instead of ~150. The log message even reports `pendingDbChanges.Count` and says the rows "will be retried at the end of the run", when in fact they are retried immediately on the next file.

`DbFailureNeverCrashesTheRun` uses a single file, so the mid-loop flush never fires — this path is untested.

#### 2.3 One locked or vanished file aborts the entire preview — [confirmed]

[`LoraSorterViewModel.cs:489`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L489)

`SorterMetadataResolver`'s XML doc promises "never throws", but its filter is `when (ex is HttpRequestException or TaskCanceledException)` — and line 66 calls `_hashFile(filePath)` → `File.OpenRead` (`FileShare.Read`), which throws `IOException`/`UnauthorizedAccessException` for a `.safetensors` currently open in ComfyUI/A1111, or on a denied ACL. `new FileInfo(path).Length` at line 490 throws `FileNotFoundException` if the file vanished between enumeration and use. And a Civitai shape change makes `DeserializeOrThrow` raise a `JsonException` that also escapes — this repo has already been bitten by that twice (`allowCommercialUse` string→array, null stats counters).

All three unwind out of `ResolveCandidatesAsync` → `Task.Run` → the catch at 433. One bad file out of 3000 kills the whole preview with zero candidates — the exact failure `EnumerateFilesSafe` was written to prevent at folder granularity.

#### 2.4 Collision hashing dropped the guard the downloader carries — [confirmed]

[`LoraSortPlanner.cs:65`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortPlanner.cs#L65)

`HashOfFile(Path.Combine(targetDir, fileName))` on a file held open by a running backend throws `IOException` out of `BuildPlan`, through the `Task.Run` at line 398, into "Preview failed" — no plan at all. `CivitaiDownloadQueue.ResolveCollisionFreeTargetPathAsync` handles precisely this with `catch (IOException) { /* Unreadable/locked — can't prove it's ours, so don't overwrite it. */ }`; the sorter dropped it while explicitly claiming to mirror the same convention.

Related: that `Task.Run` takes no `CancellationToken`, and on a cache-hit pass `_resolveCts` is null, so the overlay's Cancel button is a no-op for the entire planning phase — which can re-hash unbounded amounts of data, since `hashCache` is a per-call local and every option toggle re-plans.

---

## 3. Path and platform assumptions

#### 3.1 `IsWithin` returns false for every path when the root is a drive root — [confirmed, verified empirically]

[`LoraSorterViewModel.cs:745`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L745)

`Path.TrimEndingDirectorySeparator` deliberately does **not** trim a root path — its implementation is `EndsInDirectorySeparator(path) && !IsRoot(path) ? Substring(…) : path`. Verified with `dotnet`: `TrimEndingDirectorySeparator(GetFullPath("E:\\"))` returns `E:\`, length 3. So for `E:\Loras\a.safetensors` the boundary check reads `normalizedPath[3]` — which is `L`, not a separator — and returns false.

A dedicated LoRA drive `E:\` as source (a normal setup, and reachable via "Browse any folder") makes `ResolveCandidatesAsync` discard **every** DB-known file at line 464. All of them fall into the unknown branch: a full-file SHA256 plus a Civitai API round-trip each, thousands of needless hashes and HTTP calls — and every file loses its DB category and lands in `Unknown\`. The same helper drives the "target is another LoRA source" warning at line 424, which silently stops firing for drive-root sources.

`ModelFileSyncService.MatchEnabledRoot` gets this right with `TrimEnd(Path.DirectorySeparatorChar, …)`.

#### 3.2 UNC paths throw out of the disk pre-flight — [confirmed, verified empirically]

[`LoraSorterViewModel.cs:406`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L406)

Verified with `dotnet`: `new DriveInfo(Path.GetPathRoot(@"\\nas\share\loras"))` throws `ArgumentException: Drive name must be a root directory (i.e. 'C:\') or a drive letter`.

Pick a NAS folder as source (default target = same as source) or via "Browse target" and line 406 throws — *after* the tree was painted at 401–404 but *before* `HasEnoughSpace`/`DiskSummary`/`BlockReason` are set. The user gets a populated preview, `Preview failed: Drive name must be a root directory…`, `HasEnoughSpace` stuck at its previous value (false on a fresh tab), and Start permanently disabled with no stated reason. This PR is the first caller to feed an arbitrary user-picked folder into `GetAvailableSpace`.

#### 3.3 The 1 GB margin is applied to zero-byte same-volume moves — [confirmed]

[`LoraSorterViewModel.cs:407`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L407)

`LoraSortPlanner` correctly sets `requiredBytes = 0` for a same-volume move (line 85), but `HasEnoughSpace = free >= 0 + SafetyMarginBytes` still demands 1 GB. On the near-full drives model libraries actually live on, the *primary* use case — reorganizing LoRAs in place on the same disk, which is pure directory-entry renames — reports "Not enough free space on the target drive" and disables Start. The margin should be conditional on `RequiredBytes > 0`.

#### 3.4 Hand-rolled directory walk has no cycle guard — [confirmed]

[`LoraSorterViewModel.cs:503`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L503)

38 lines of stack-based recursion where `EnumerationOptions { RecurseSubdirectories = true, IgnoreInaccessible = true }` does the job in three — and the repo already declares exactly that as `RecursiveSafe` in [`ImageResourceHasher.cs:31`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Distiller/ImageResourceHasher.cs#L31) and [`PipelineAssetInstaller.cs:35`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Pipelines/PipelineAssetInstaller.cs#L35).

The hand-rolled version is behaviourally worse in three ways: a directory junction pointing at itself or an ancestor (`mklink /J`, or a symlinked shared-models dir — common in multi-backend setups) grows the pending `Stack` without bound and the preview never returns; when `Directory.GetFiles(dir)` throws at line 517 the handler `continue`s *before* calling `GetDirectories`, so that subtree is dropped entirely rather than partially enumerated; and `DirectoryNotFoundException` from a mid-walk race is not in the filter, so it escapes and kills the preview — the exact failure the doc comment says the method exists to prevent.

Worth noting the one test that covers this, `InaccessibleSubfolderIsSkippedNotFatal`, states in its own comment that it does not create an inaccessible folder — so the fallback is currently untested.

#### 3.5 Windows-only path logic without the required Linux marker — [plausible]

[`LoraSortPlanner.cs:82`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortPlanner.cs#L82)

`.github/copilot-instructions.md` requires a `TODO: Linux Implementation` comment for Windows-only work, and that the code be left open to extension. The diff contains zero `TODO` comments across 3117 lines.

Two spots matter. `Path.GetPathRoot` returns `/` for every absolute path on Linux, so `sameVolumeMove` is always true, `requiredBytes` collapses to 0, and the free-space pre-flight silently passes for any cross-device move. And `SorterPathBuilder.SanitizeFolderName`'s `TrimEnd('.', ' ')` is a Win32-only restriction while `Path.GetInvalidFileNameChars()` returns only `\0` and `/` on Linux, so `\ : * ? " < > |` would pass straight into created directory names. Both are `static` with no injected policy, so neither is currently open to extension.

---

## 4. Parity with the download pipeline

This is the cluster I would most want a decision on before merge, because two of these already diverge *today* — the comments assert a parity the code does not have.

#### 4.1 Sorted and downloaded files do **not** land in identical folders — [confirmed]

[`SorterCategoryResolver.cs:13`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/SorterCategoryResolver.cs#L13)

The spec (`docs/superpowers/specs/2026-08-20-lora-sorter-design.md:64`) asked for the download pipeline's category logic to be *extracted into one shared helper*. This adds a fifth copy instead — the enum-returning version already exists essentially verbatim as `ModelDetailViewModel.InferCategoryEnumFromTags`, with string twins in `ModelDetailViewModel`, `CivitaiResultViewModel`, `DownloadLoraDialogViewModel` and `MetaDataUtilService`.

More importantly the results already differ. When no tag matches, the downloader returns `null` and `DownloadDestinationViewModel.BuildTargetDirectory:124` **skips the category segment** (`<root>\SDXL 1.0\file.safetensors`), while the sorter returns `Unknown` and `SorterPathBuilder` appends it unconditionally (`<root>\SDXL 1.0\Unknown\file.safetensors`). Same divergence for a null base model — the sorter writes `Unknown\`, `DownloadDestinationViewModel` omits the segment, and `CivitaiDownloadQueue:521` uses a third name, `Unsorted`.

Both features default their folder toggles to `true`, so this is the common path, not an edge case: **every sort run relocates uncategorized downloads into `Unknown\`, and the next download re-creates them one level up.** Permanent ping-pong, and a user ends up with `Unknown\`, `Unsorted\` and loose files simultaneously.

#### 4.2 Browsed-folder files are forced to `Unknown` even with full metadata — [confirmed]

[`LoraSorterViewModel.cs:492`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L492)

`SorterMetadataResolver` correctly returns `BaseModelRaw` and `CivitaiVersionId` from a `.civitai.info` sidecar or the API, and then line 492 passes `SorterCategoryResolver.ToFolderName(CivitaiCategory.Unknown)` unconditionally. With "Base model + category" selected, a folder of properly downloaded LoRAs lands entirely in `<Target>\SDXL 1.0\Unknown\` instead of `\Character\`, `\Style\`, etc.

This is the *normal* path for the headline "Browse any folder" feature, because `LoadCachedFilesAsync` hard-filters to enabled LoRA source roots (`MatchEnabledRoot`, and it returns `[]` outright when no source is enabled) — so an unregistered folder gets no DB metadata at all. The spec reserves `Unknown\Unknown` for files that are "still unresolved"; here it is applied to fully resolved ones. `UnknownFileInBrowsedFolderIsResolvedIntoUnknownBuckets` only covers the genuinely-unresolvable case, so the gap is invisible in the suite.

Also worth flagging: `LoraViewer.md` documents a resolution step that does not exist — *"DB lookup by path, then by SHA256 hash (file known under another path)"*. `SorterMetadataResolver` does sidecar → disk cache → API, never a DB hash lookup. That step was descoped by ruling; the doc just has not caught up.

#### 4.3 Copy-mode re-runs are not idempotent — [confirmed]

[`SorterPathBuilder.cs:51`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/SorterPathBuilder.cs#L51)

`BuildCollisionFreeFileName` never content-compares the version-id-suffixed candidate. Copy mode, target ≠ source: run 1 copies `V1.safetensors` (versionId 42, content X) in as `V1_42.safetensors` because the plain name is held by different content Y. Run 2 — source untouched, since copy mode does not move it — collides on the plain name again, finds `nameIsTaken("V1_42.safetensors")` now true, falls through to `V1_2.safetensors` and copies X a second time. Run 3 → `V1_3`, run 4 → `V1_4`, unbounded.

The header comment's "so re-runs are idempotent" and the checklist's "Re-run same sort: zero transfers" both fail in copy mode. `CivitaiDownloadQueue.ResolveCollisionFreeTargetPathAsync` avoids this by hash-comparing and returning the suffixed path unconditionally — it has no `_2`/`_3` tail at all, which also means a sorter-produced `_N` name is one the downloader will never recognise as its own.

#### 4.4 The sidecar list is narrower than the canonical one — [plausible]

[`SidecarLocator.cs:15`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/SidecarLocator.cs#L15)

`SidecarExtensions` is a hand-copy of `StaticFileTypes.GeneralExtensions` missing every video extension (`.mp4`, `.webm`, `.mov`, `.avi`, `.mkv`, `.wmv`, `.flv`, `.m4v`) plus `.civitai` and `.thumb`. A LoRA with a video preview gets its weights moved and `MyLora.mp4` left behind — permanently detached, and with `DeleteEmptySourceFolders` on the folder is not even cleaned up because the orphan keeps it non-empty. `ModelDiscoveryService` treats those files as part of the model, so the two views of "what belongs to this model" now disagree.

Related: `DeriveSidecarTargetPath`'s `sidecarName[sourceStem.Length..]` slice is unguarded and only safe under `FindSidecars`' unwritten invariant. `ModelMetadataUtils.ExtractBaseName` already solves multi-dot stems correctly by longest-known-suffix match. (`StaticFileTypes` being `internal` to Service means sharing it needs it public or moved to Domain — a real cost, but the drift is already live.)

#### 4.5 `Enum.TryParse` accepts numeric and comma-separated tags — [plausible]

[`SorterCategoryResolver.cs:24`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/SorterCategoryResolver.cs#L24)

Against the real enum (`Unknown=0, Character=1, Style=2, Celebrity=3, Concept=4, Clothing=5`): tag `"5"` parses to `Clothing`; tag `"2000"` — a real Civitai style tag — parses to the *undefined* value 2000, and `ToFolderName` returns `category.ToString()`, so the sorter creates and populates a directory literally named `2000`; tag `"character,style"` parses to `1|2 == Celebrity`. None hit the `category != CivitaiCategory.Unknown` guard.

This is inherited from the original, but the original was display-only — here the value names a directory files are physically moved into. An `Enum.IsDefined` check (plus rejecting leading digits) closes it.

#### 4.6 Stored DB hashes are compared without normalization — [plausible]

[`LoraSortPlanner.cs:27`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortPlanner.cs#L27)

`LoraDuplicateFinder.NormalizeHash` exists because *"ModelFile.HashSHA256 may have been stored in mixed case or with separators from older import paths"*. `HashOfCandidate` trusts `c.Sha256` verbatim, so a legacy dashed hash never equals a freshly computed one, `PlannedAction.SkippedDuplicate` is never reached, and the identical file is renamed and transferred — producing exactly the duplicate the collision policy promises to skip. `NormalizeHash` also rejects stored values whose stripped length is not 64; the sorter accepts a truncated or garbage DB hash as content identity.

---

## 5. Lifetime and threading

#### 5.1 A long-lived shared `DbContext` is used from overlapping preview passes — [plausible]

[`LoraSorterViewModel.cs:449`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L449)

`IModelSyncService` and `IUnitOfWork` are transient, so the scoped `LoraViewerViewModel` holds one `DbContext` for the session and forwards that same instance to the sorter (line 337). `RecomputePreviewCoreAsync` is invoked fire-and-forget from three option hooks and `RunBusyAsync` has no re-entrancy guard, so: switch source folder (pass A enters `Task.Run` → `LoadCachedFilesAsync`, a multi-second `AsSplitQuery` over a large library), then toggle "Base model + category" → pass B misses the still-unpopulated cache and issues a second query on the same context → `InvalidOperationException: A second operation was started on this context instance before a previous operation completed`.

Every other consumer resolves a fresh scope for exactly this call — `LoadCachedTilesAsync` (:439), `DownloadMissingMetadataAsync` (:685), `RebuildTilesFromDatabaseAsync` (:905). This is also `_syncService`'s first direct dereference anywhere in `LoraViewerViewModel`; it was previously only null-checked.

Related: a superseded pass's `RunBusyAsync` `finally` tears down the busy overlay while the newer pass is still resolving, which is one of the ways to reach 1.1.

#### 5.2 The design-time constructor fires a real preview that wipes the demo data — [plausible]

[`LoraSorterViewModel.cs:89`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L89)

`SelectedSourceFolder = SourceFolders[0]` fires `OnSelectedSourceFolderChanged` with `_isInitializing` still `false`, so `_ = RecomputePreviewAsync()` starts real disk and `DriveInfo` work against `C:\Demo\Loras`. The ctor then adds the two demo `PreviewRoots` and `TransferCount = 17`; the continuation resumes, `BuildPreviewTree` calls `PreviewRoots.Clear()` and sets `TransferCount = 0` — so the previewer shows an empty tree, the opposite of what the demo data is for.

`LoraViewerViewModel`'s design-time ctor now constructs this VM too, so the five existing `new LoraViewerViewModel()` calls in `LoraViewerFilterPersistenceTests` each spawn background filesystem I/O off the test thread, with any exception unobserved. Setting `_isInitializing = true` in that ctor (or assigning the backing field) fixes it.

---

## 6. Performance

#### 6.1 The whole manifest is rewritten once per transferred file — [confirmed]

[`SortHistoryWriter.cs:42`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/SortHistoryWriter.cs#L42)

`MarkCompleted` does `ReadAllText` + `Deserialize` + a full `Select` rebuild + indented `Serialize` + `WriteAllText`, from inside the per-file loop. A 3000-LoRA sort means roughly 9M entry (de)serializations and 6000 whole-file I/O operations on a multi-MB JSON file, on the transfer hot path — O(n²) in library size.

The durability angle matters more than the cost: `File.WriteAllText` truncates before writing, so a kill or power loss during any of those n writes leaves truncated JSON — destroying the crash-recovery artifact the file exists to be, and making the *next* run's `Deserialize<Manifest>(...)!` throw the `JsonException` from 2.1. Appending one line to a JSONL completion log, or flushing the completed set on the DB-batch cadence, is O(N) and equally restorable.

Two smaller notes on the same file: `WritePlan` (line 37) is called outside any try/catch, so a failure writing this currently-unread side-file aborts the sort before a single file is touched; and the manifest schema does not record sidecar targets at all, so a Restore built on it would move weights back and strand every companion file.

#### 6.2 The sorter initializes on LoRA Viewer load, not on tab open — [confirmed]

[`LoraSorterView.axaml.cs:39`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Views/LoraSorterView.axaml.cs#L39)

Initialization triggers from `OnDataContextChanged` as well as `OnAttachedToVisualTree`. The view is inline XAML inside a `TabItem`, and DataContext inheritance flows through Avalonia's logical tree regardless of tab selection — so the binding resolves, `_initialized` flips, and `InitializeAsync()` runs on every LoRA Viewer open even if the user never touches the Sorter tab: a full disk walk, SHA256 of every unknown file, and by-hash API requests. `CivitaiBrowserView`, the sibling tab, has no attach-time initialization. Gating on `OnAttachedToVisualTree` alone (or on the `TabControl`'s `SelectedItem`) keeps it lazy.

#### 6.3 The API key is re-read through a fresh DI scope for every file — [plausible]

[`SorterMetadataResolver.cs:77`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/SorterMetadataResolver.cs#L77)

`await _apiKeyProvider()` is inline in the API call. In production that is `GetApiKeyForSorterAsync`, which creates an `IServiceScope`, resolves `IAppSettingsService` and queries the DB. A folder of 1000 unresolved LoRAs means 1000 scopes, 1000 `DbContext`s and 1000 queries for a value that cannot change during the run. The codebase's own bulk equivalent, `DownloadMissingMetadataAsync` (`LoraViewerViewModel.cs:659`), deliberately reads the key once before its loop.

The surrounding loop is also fully sequential — full-file SHA256 plus a serialized round-trip per file — and repeats after every sort run, since line 608 nulls the candidate cache. A bounded `Parallel.ForEachAsync` would cut a multi-minute pass down substantially.

---

## 7. Reuse and conventions

#### 7.1 Another private `FormatBytes` — [confirmed]

[`SortPreviewNodeViewModel.cs:25`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/SortPreviewNodeViewModel.cs#L25)

`DiffusionNexus.UI/Helpers/FileSizeFormatter.cs` was extracted during the #442 housekeeping pass specifically to consolidate these — its own doc comment says so. The new copy uses `:F1` where `FileSizeFormatter` and `CivitaiDownloadQueue` use `:F2`, so the sorter's disk gate reads "4.2 GB" while the download queue's gate for the same bytes reads "4.20 GB", on the same screen. Two unrelated call sites (`LoraSorterViewModel.cs:408`, `:579`) also now reach into a tree-node VM's `internal static` method purely to format a number.

Same shape at [`LoraSorterViewModel.cs:751`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L751): `ComputeSha256` sits next to `HashingService` and `LoraViewerViewModel.ComputeFullSha256` — the latter being what this PR already injects as the runtime `hashFile` delegate, making the private copy reachable only from the design-time ctor.

#### 7.2 Happy-path logging is missing, and the doc claims otherwise — [plausible]

[`LoraSortExecutor.cs:95`](https://github.com/Little-God1983/DiffusionNexus/blob/feature/lora-sorter/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortExecutor.cs#L95)

The standing project rule is that every new feature instruments its working steps to the Unified Console — not just errors — with timings, so a hang shows the last step that succeeded. Around `await _pathUpdater.UpdateLocalPathsAsync(...)` the only `_logger` calls are two `Error` calls, and `DbLocalPathUpdater` takes no `IUnifiedLogger` at all. The slowest loop in the feature (`LoraSorterViewModel.cs:489`) reports only into `BusyMessage`, never the console, and the diff contains no `Stopwatch` anywhere — so "slow" and "hung" are indistinguishable from an exported log.

`DiffusionNexus.UI/Doc/LoraViewer.md:533` asserts the opposite: *"every step logs to the Unified Console … plan summary, each file's transfer, DB batches, final tally."* Either the logging or the doc should move.

---

<sub>Review-only pass — no code was changed. `[plausible]` findings were not independently re-verified; worth confirming before acting on them.</sub>

Little-God1983:
## Review response — all 24 findings addressed

Every finding was verified against the code (and empirically where the claim was about runtime behavior) before any change. 27 commits on the branch; full test project **3967 green**, UI Release build clean. Fixed one finding at a time, each with its own tests.

### Fixed as described
| # | Fix |
|---|---|
| 1.1 | Failed/cancelled preview now **disarms** the plan (`DisarmPlan`: `_lastPlan=null`, counts/gate/tree reset, reason in `BlockReason`); CTS is owned per pass (also closes a cross-pass `ObjectDisposedException` found along the way). |
| 1.2 | Empty-folder cleanup is behind a seam, wrapped; failure → Warn + note in the result; recompute + `SortCompleted` still run. |
| 2.1 | Terminal DB flush is in `finally`; `MarkCompleted` wrapped; sidecar-slice guarded. |
| 2.2 | Failed batch moves to a `deferred` list retried **once** at the end; pending is cleared; the triggering file still gets MarkCompleted/done/progress. Test with 21 transfers + failing updater. |
| 2.3 | Per-file guard in candidate resolution (`IOException`/`UnauthorizedAccessException`/`JsonException`/`ArgumentException`) → skip + "N files skipped" in the status bar; resolver's hash read + API deserialization guarded too. |
| 2.4 | Unreadable hash during collision check ⇒ "can't prove identical ⇒ rename" (the downloader's guard); `BuildPlan` takes the pass token so Cancel works during hashing. |
| 3.1 | `IsWithin` rewritten (root normalized to end in exactly one separator); 9-case theory incl. `E:\` root. Confirmed the `TrimEndingDirectorySeparator("E:\")` behavior by probe. |
| 3.2 | UNC/unmapped free-space probe **fails open** with "Free space unknown (network or unsupported path)". |
| 3.3 | 1 GB margin applied only when `RequiredBytes > 0`. |
| 3.4 | Walk replaced with `EnumerationOptions { RecurseSubdirectories, IgnoreInaccessible, AttributesToSkip = ReparsePoint | System }` — reparse-point skip is the cycle guard (junction cycle reproduced by probe). |
| 3.5 | `TODO: Linux Implementation` markers at the `GetPathRoot` same-volume check and the Win32 trailing-dot/space rule. |
| 4.1 | **Decision: downloader parity.** Unknown category ⇒ segment omitted (`<root>\<BaseModel>\file`), exactly like `BuildTargetDirectory`; base-model `Unknown\` folder stays for placeholder base models. Category inference is now **one shared helper** (`SorterCategoryResolver.InferFolderName`) that `CivitaiResultViewModel` and `DownloadLoraDialogViewModel` delegate to — parity by construction. |
| 4.2 | Sidecar `tags` (when present) feed category inference; `LoraViewer.md` no longer claims a DB-by-hash step (descoped by ruling). |
| 4.3 | Copy-mode re-runs idempotent: a taken version-suffixed name is hash-compared → identical ⇒ `SkippedDuplicate`. Test: run 2 in copy mode ⇒ zero transfers. |
| 4.4 | Sidecar list now a superset of `StaticFileTypes.GeneralExtensions` minus model extensions (video previews, `.civitai`, `.thumb`). |
| 4.5 | Numeric tags rejected — see correction below. |
| 4.6 | Stored hashes normalized (separators stripped, exactly 64 hex required, else lazy-hash). |
| 5.1 | `LoadCachedFiles` goes through a fresh DI scope per pass (lifetimes confirmed: `IUnitOfWork`/`IModelSyncService` transient, `LoraViewerViewModel` scoped). |
| 5.2 | Design-time ctor no longer triggers a real preview. |
| 6.1 | Manifest: plan JSON written once + **append-only completion JSONL**; sidecar source/target recorded for a future Restore; `WritePlan` failure is non-fatal. |
| 6.2 | Init only from `OnAttachedToVisualTree` (tab open), not `OnDataContextChanged`. |
| 6.3 | API key memoized per resolution pass (`ResetApiKeyCache()`). Parallelism declined (YAGNI). |
| 7.1 | `FileSizeFormatter` + `HashingService` reused; private copies deleted. |
| 7.2 | `Stopwatch`-timed Info lines for resolution / plan / DB batches / run; doc now matches code. |

### Claims corrected on evidence
- **4.5's proposed fix was insufficient**: `"character,style"` parses to the *defined* value `Celebrity` (1|2), so `Enum.IsDefined` + leading-digit rejection doesn't close it — a comma check was required and added. Also applied `Enum.IsDefined` to the stored `UserCategory` override.
- **2.1's `ArgumentOutOfRangeException`** from the sidecar slice is unreachable via `FindSidecars` (both sides derive the stem from the same input) — latent, guarded anyway.
- **4.6's "mixed case" half** was already handled by the `OrdinalIgnoreCase` compare; only separators and length validation were live.
- **4.1's "permanent ping-pong"** is one-directional (downloads land loose, the sorter tidies them once) — but the parity violation itself was real, hence the decision above.
- **4.2 "forced to Unknown even with full metadata"**: neither the by-hash API nor standard `.civitai.info` files carry tags, so there was no category to infer; tags are now used when a sidecar does carry them.
- **6.2's `TabControl.SelectedItem` alternative** is unnecessary — attach already implies selection. **5.2's delay-based assertion** would be racy — a synchronous `IsBusy` assertion is used instead.

### Still owed
Manual GUI smoke (lazy tab-open init, the new status strings, a run against a drive-root and a UNC source, multi-GB copy responsiveness).

Little-God1983:
## Verification review — the fix wave (`f482dcf..9863678`)

Follow-up to [the 24-finding review](https://github.com/Little-God1983/DiffusionNexus/pull/520#issuecomment-5361563781) and [the response claiming all 24 addressed](https://github.com/Little-God1983/DiffusionNexus/pull/520#issuecomment-5362847729). This pass checked, for each finding, whether the fix eliminates the **failure scenario** or only the example used to illustrate it — and whether the fix introduced anything new. Solution builds clean, 0 warnings. Line links point at `9863678` so they stay valid if the branch moves.

**Result: 20 of 24 genuinely fixed, 3 partial, 4 new regressions. 13 findings below.**

First, credit where it is due — most of this wave is real engineering, not symptom-patching. 2.1's `try`/`finally` + `MarkCompletedSafely`, 2.2's `deferredDbChanges` folded into the terminal flush exactly once, 3.1's `IsWithin` (correct against all nine boundary cases including `E:\`), 4.6's `NormalizeHash`, 5.1's fresh scope with `AsNoTracking()` + eager `Include`s so post-dispose access is safe, and 6.1's append-only JSONL all address the underlying cause. Several arrived with tests that exercise the actual failure rather than the happy path; the junction test in particular proves `AttributesToSkip` blocks *recursion*, not merely inclusion, which is the part that mattered.

### Per-finding verdicts

| # | Verdict | Note |
|---|---|---|
| 1.1 | **PARTIAL** | `DisarmPlan` covers the failed/cancelled path; the success path has no ownership guard and `IsBusy` is still shared — **A1** |
| 1.2 | FIXED | guarded; `_emptyFolderCleanupFailed` surfaced; cache-invalidate + `SortCompleted` always run |
| 2.1 | FIXED | loop in `try`/`finally`, `MarkCompletedSafely` outside the narrow filter |
| 2.2 | FIXED | `deferredDbChanges` cleared and flushed once |
| 2.3 | FIXED | per-file guard in the VM, `_hashFile`/`JsonException` guards in the resolver — but see **B2**, **C2** |
| 2.4 | FIXED | `catch (IOException)` around hashing, `CancellationToken` threaded into `BuildPlan`, CTS now created for cache-hit passes |
| 3.1 | FIXED | verified against all nine boundary cases incl. `E:\` |
| 3.2 | **FIXED, over-corrected** | UNC now fails open — and so does a genuinely dead drive letter — **B1** |
| 3.3 | FIXED | `RequiredBytes == 0` ⇒ no margin |
| 3.4 | FIXED | real `EnumerationOptions`; but see **B3** and **D1** |
| 3.5 | FIXED | two accurate `TODO: Linux Implementation` blocks |
| 4.1 | FIXED | both downloader VMs delegate; `Unknown` contributes no segment on either side |
| 4.2 | **PARTIAL** | sidecar tags only; cache and API paths still land category-less — **A2** |
| 4.3 | **PARTIAL** | only the `_{versionId}` slot is content-compared — **A3** |
| 4.4 | FIXED | list verified as a superset of `GeneralExtensions`; slice guarded |
| 4.5 | FIXED | `LooksLikeCategoryName` + `Enum.IsDefined` on both the tag and override paths |
| 4.6 | FIXED | separators stripped, lowercased, non-64-hex rejected |
| 5.1 | FIXED | fresh scope; `AsNoTracking()` + eager includes make post-dispose access safe |
| 5.2 | FIXED | `_isInitializing` latched, demo data intact |
| 6.1 | FIXED | append-only JSONL, `WritePlan` fails soft — but see **D2** |
| 6.2 | FIXED | inline `TabItem` content confirmed; attach-only is the right gate |
| 6.3 | FIXED | memoized key + `ResetApiKeyCache()` per pass |
| 7.1 | FIXED | shared `FileSizeFormatter` + `HashingService` |
| 7.2 | FIXED | timings throughout, asserted by test |

`933f9ad`/`6b4d197` are accurate doc updates. **`9863678` is the weakest commit in the wave** — see **C1** and **C3**.

---

## A. Findings that are not actually closed

#### A1 — 1.1: the stale-armed-Start window survives

[`LoraSorterViewModel.cs:456`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L456)

`DisarmPlan` correctly handles a pass that fails or is cancelled. Two gaps remain.

The **success** path never calls `IsCurrentPass()`, and the commit guard runs only in the cache-miss branch and compares only source/target folders — never `IncludeCategory`/`IsMove`. So if a superseded pass's `Task.Run` happens to finish *after* the pass that replaced it, it resumes on the success path and permanently overwrites the newer plan and preview tree.

Separately, `RunBusyAsync`'s `IsBusy` is shared across overlapping passes, so the first one to finish drops the overlay while another is still computing. Pass A succeeds (42 files, Move). Toggling `IncludeCategory` starts B; toggling `IsMove` starts C and cancels B. B observes cancellation, returns at `if (!IsCurrentPass()) return;`, and its `finally` sets `IsBusy = false` while C is still running — so `CanStart` goes true against **A's** `_lastPlan`/`TransferCount`/`HasEnoughSpace`. Start then executes A's Move plan while the radio reads Copy, and `ReferenceEquals(_lastPlan, plan)` cannot see it because nothing reassigned `_lastPlan`.

A pass-ownership check on the success path, plus per-pass busy tracking (or including the options in the commit guard), closes both.

#### A2 — 4.2: only sidecar-sourced tags carry a category

[`SorterMetadataResolver.cs:102`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/Services/Lora/Sorting/SorterMetadataResolver.cs#L102)

`TryReadSidecar` now propagates tags — but the other two resolution paths do not. `TryReadCache` returns `cached! with { Sha256 = sha }`, and `CacheEntry` carries only `BaseModel`/`VersionId`; the API branch returns `new ResolvedLoraMetadata(version.BaseModel, version.Id, sha)`. `Tags` is `[]` in both.

So for the headline case — browsing a folder of LoRAs downloaded outside DiffusionNexus, with no `.civitai.info` alongside them — `InferFolderName([])` returns null, category falls back to `UnknownFolderName`, and `BuildTargetDirectory` drops the segment. Every file lands flat in `<Target>\<BaseModel>\`. That is the same defect 4.2 described; it is just no longer visible as an `Unknown\` folder. The per-hash cache makes it sticky: a second preview of the same folder cannot recover even if the by-hash DTO later gained tags.

The commit subject ("browsed-folder files use their **sidecar** tags") is accurate. The claim that finding 4.2 is addressed is not.

#### A3 — 4.3: the numeric fallback is still never content-compared

[`LoraSortPlanner.cs:110`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/Services/Lora/Sorting/LoraSortPlanner.cs#L110)

The hash comparison was added to the `_{versionId}` slot only. `BuildCollisionFreeFileName`'s `_2`/`_3`/`_4`… fallback still receives no content check.

**Case A — no version id at all**, which is exactly what the numeric convention exists for. `V1.safetensors`, `CivitaiVersionId == null`, content A; target already holds `V1.safetensors` with content B. Run 1 copies it in as `V1_2.safetensors`. Run 2: plain name still taken, `if (candidate.CivitaiVersionId is { } versionId)` is skipped entirely, and `V1_2` is seen as "taken" — by its own run-1 copy, never hashed — so it yields `V1_3`. Run 3 yields `V1_4`. Unbounded, identical to the original bug.

**Case B** — version id present but its slot holds a third model. The wave's own test `DifferentContentAtTheVersionSuffixedNameStillFallsToNumericSuffix` lands on `V1_2`; re-running that same scenario produces `V1_3`.

Both new tests use `versionId = 42` with the suffixed slot holding *our own* copy, so neither case is covered.

---

## B. Regressions introduced by the wave

#### B1 — 3.2 over-corrected: a dead drive letter now enables Start

[`LoraSorterViewModel.cs:944`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L944)

`ApplyDiskPreflight` fails open for *every* probe failure. `DriveNotFoundException` derives from `IOException`, so a target on an unplugged `Z:\` gets `HasEnoughSpace = true`, `BlockReason = null`, and "Free space unknown (network or unsupported path)". Start is enabled, and the executor then throws on `CreateDirectory` for every single file: *"Done: 0 sorted, 0 duplicates skipped, 412 failed."*

UNC genuinely warrants failing open — `new DriveInfo(@"\\nas\share\")` throws `ArgumentException`, and there is no answer to give. "Drive does not exist" is a different thing and is distinguishable by exception type, or by a `Directory.Exists(targetRoot)` probe. `UnknowableFreeSpaceFailsOpenInsteadOfBlockingTheRun` currently locks the wrong behaviour in for `DriveNotFoundException`.

#### B2 — 2.3: `knownPaths.Add` moved, so failed known files get re-resolved as unknown

[`LoraSorterViewModel.cs:606`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L606)

The `Add` used to be the first statement in the try; it is now second-to-last, after the candidate is constructed. So when `new FileInfo(path).Length` or `SidecarLocator.FindSidecars` throws — a DB-known `.safetensors` held open by a running ComfyUI — the path is counted in `skipped++` but never registered as known.

`EnumerateModelFilesSafe` then yields it again, `!knownPaths.Contains(...)` passes, and the unknown loop runs a full-file SHA256 plus a serialized Civitai by-hash round-trip on the same file. If it fails again the user is told "2 file(s) skipped" for one file. If it succeeds, the file is planned from API metadata instead of its own DB row — losing `UserCategory`, the DB base model, and the stored hash. Moving the `Add` back above the size/category work restores the invariant.

#### B3 — 3.4: `System` attribute filter silently drops model files

[`LoraSorterViewModel.cs:720`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L720)

`AttributesToSkip = ReparsePoint | System`. The replaced walk used `Directory.GetFiles`/`GetDirectories`, which return entries regardless of `Hidden` or `System`. Clearing the default `Hidden` filter was right; adding `System` is a behaviour change, and `EnumerationOptions` applies it to files as well as directories.

A `.safetensors` with the System bit set — common on files restored by a backup tool, copied off a NAS, or under a folder a sync client marked System — simply does not appear in the preview and is never sorted. No warning, no log line, and the "N file(s) skipped" note stays at 0 because nothing threw. `ReparsePoint` alone is what the documented cycle guard needs. This is also absent from the limitations list the same wave added to `LoraViewer.md`.

#### B4 — 1.1: Cancel wipes the preview tree, then says it didn't

[`LoraSorterViewModel.cs:508`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L508)

With a good 42-file preview on screen, toggling an option and pressing Cancel on the busy overlay runs `DisarmPlan(...)` — `PreviewRoots.Clear()`, `PreviewSummary = null`, `DiskSummary = null`, `TransferCount = 0` — and then sets `StatusMessage` to *"Cancelled — preview not updated."*

The user sees an empty tree, a zeroed summary, and a message asserting nothing changed, alongside a `BlockReason` ("press Refresh to rebuild it") that the status line immediately contradicts. Disarming `_lastPlan`/`TransferCount`/`HasEnoughSpace` is correct and necessary; clearing the rendered tree is not. The cancel path should keep the tree and mark it stale.

---

## C. New defects in the fix code

#### C1 — `9863678` widened a blanket catch instead of fixing the cause

[`LoraSorterViewModel.cs:698`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L698)

To handle a whitespace-only `LocalPath`, `ArgumentException` was added to `IsSkippableFileFailure` — which also admits `ArgumentNullException` and `ArgumentOutOfRangeException`. The unknown-file loop wraps `await _metadataResolver.ResolveAsync(path, ct)` in that filter, so an `ArgumentNullException` raised anywhere inside the resolver, `ICivitaiClient`, `SidecarLocator` or `SorterCategoryResolver` — an NRE-class programming defect — is now counted as one "locked/unreadable" file and reported to the user as such, behind a single `Warn`.

A systematic `ArgumentException` (a bad cache directory, say) on a 3000-file folder reports *"3000 file(s) skipped (locked/unreadable)"* with an empty preview, pointing the user at their disk instead of at the bug. `if (string.IsNullOrWhiteSpace(path)) continue;` is the one-line fix, and it also removes the need to run `IsWithin`/`_fileExistsOnDisk` inside the try at all.

#### C2 — user cancellation is logged as a network failure

[`SorterMetadataResolver.cs:117`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/Services/Lora/Sorting/SorterMetadataResolver.cs#L117)

The filter is `when (ex is HttpRequestException or TaskCanceledException or JsonException)`. `TaskCanceledException` derives from `OperationCanceledException` — the very type the 2.3 fix documents one level up as *"deliberately not covered: cancellation must still unwind the pass."*

Press Cancel during a 1000-file resolve and the in-flight request faults with `TaskCanceledException`, which this filter swallows: the pass logs *"Civitai by-hash lookup failed for `<path>`: A task was canceled"* and returns unresolved metadata for a file that was never looked up. Only the next iteration's `ct.ThrowIfCancellationRequested()` stops the loop, so the cancel lands one file late and appears in the exported console as a network failure. `ex is TaskCanceledException && !ct.IsCancellationRequested` separates a genuine HTTP timeout from a user cancel.

#### C3 — `9863678` half-applied its own insight

[`LoraSorterViewModel.cs:886`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L886)

The commit switched empty-folder cleanup to `plan.SourceRoot`, reasoning that it should use "the plan that actually ran, not the possibly since-changed live selection". But the gate above it still reads live UI state: `if (IsMove && DeleteEmptySourceFolders && !result.Cancelled)`, while the run it is deciding about used `plan.IsMove` and the `DeleteEmptySourceFolders` captured at `BuildOptions()` time.

The busy overlay blocking the toggles is currently the only thing keeping this latent. `LoraSortPlan` already carries `IsMove`; `LoraSortOptions` carries `DeleteEmptySourceFolders` but is not threaded onto the plan, so that second flag has no snapshot to read.

#### C4 — TOCTOU on `_resolveCts`

[`LoraSorterViewModel.cs:751`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs#L751)

`RecomputePreviewCoreAsync`'s `finally` does `_resolveCts = null;` then `passCts.Dispose();`. `CancelSort` does `if (_resolveCts is not null) { …; _resolveCts.Cancel(); }` — two separate reads of a mutable field — and the new-pass path calls `_resolveCts?.Cancel()` on a source another pass may already be disposing.

In production this is unreachable: `RunBusyAsync` has no `ConfigureAwait(false)`, so every continuation marshals back to the UI thread. But the comment above the `Cancel` asserts *"`_resolveCts` is null whenever no pass is in flight, so `Cancel()` can never land on a disposed source"* — which is true only under a UI-thread assumption it does not state. Without a `SynchronizationContext` (the test project, or any future headless path) the continuation lands on the pool and these interleave into `NullReferenceException` or `ObjectDisposedException`. Worth either stating the assumption in the comment or taking a local copy.

---

## D. Tests and surface

#### D1 — 3.4's rewrite deleted its ACL test without replacing it

[`LoraSorterViewModelTests.cs:475`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.Tests/Sorter/LoraSorterViewModelTests.cs#L475)

`InaccessibleSubfolderIsSkippedNotFatal` was removed and nothing took its place. The replacements added in the same commit cover deep nesting and reparse points, but never an ACL denial — so the guarantee that a permission-denied subtree yields a partial preview now rests entirely on `EnumerationOptions.IgnoreInaccessible`, implicit framework behaviour with no local coverage.

To be fair, the deleted test was itself hollow — it never created an inaccessible folder, which I flagged in 3.4. That makes this a chance to write the real one rather than restore the old one.

#### D2 — `ReadCompleted`/`SidecarEntries` have no production caller

[`SortHistoryWriter.cs:101`](https://github.com/Little-God1983/DiffusionNexus/blob/9863678/DiffusionNexus.UI/Services/Lora/Sorting/SortHistoryWriter.cs#L101)

`ReadCompleted` is referenced only by its own declaration and by `LoraSortExecutorTests`/`SortHistoryWriterTests`. `SidecarEntries(move)` additionally runs `SidecarLocator.DeriveSidecarTargetPath` once per sidecar for every planned transfer at `WritePlan` time — work whose only consumer is a Restore feature the docs say is not in v1.

Recording sidecar targets is a genuine improvement over what I flagged in 6.1 (the old schema captured none, so a Restore would have stranded them). The note is narrower: a public API and a per-plan loop that only tests read, verifying a format no shipping code consumes. If Restore lands with a different schema, those tests were guarding nothing.

---

<sub>Verification pass over `f482dcf..9863678`, review-only — no code changed. Build clean; findings are from code reading plus the wave's own tests, not from running the app.</sub>

Little-God1983:
## Response to the verification review — all 13 addressed

Agreed with the assessment: the first wave closed examples in three places and regressed four. Every item below was traced as a **scenario** through the new code, fixed one commit at a time, and — where the claim was about runtime — proven by running the new test against the reverted code first.

| # | Fix |
|---|---|
| **A1** | Per-pass **generation counter** claimed *before* the predecessor is cancelled; every commit site (cache-hit AND cache-miss, success AND failure) requires `IsCurrentPass()`. Preview busy state is now an **in-flight counter** (`IsBusy = count > 0`, Interlocked, inc/dec in `try/finally`) so no overlapping pass can drop the overlay early. The null-source early-return also bumps the generation. Test: slow pass A released after B finishes ⇒ A is not committed, `IsBusy` held until both finish. |
| **A2** | After a by-hash hit, `GetModelAsync(version.ModelId)` supplies tags (`ICivitaiClient` already had it); cached per hash; "tags unresolved" is cached as *null* so a failed model lookup retries next pass instead of making the file category-less forever; a per-pass `modelId→tags` memo avoids repeated `/models/{id}` calls for versions of the same model. Old cache files without the field still parse. |
| **A3** | Collision-name selection moved into the planner: **every** name in the sequence (plain → `_{versionId}` → `_2`, `_3`…) is content-compared against both in-plan claimants and on-disk files — identical ⇒ `SkippedDuplicate`, different ⇒ next name. Tests: null-versionId copy re-run ⇒ zero transfers; version slot held by a third model, re-run ⇒ zero transfers; `_2` held by different content ⇒ `_3`. |
| **B1** | Fail-open only for `ArgumentException` **and** an existing target root; `DriveNotFoundException`/other `IOException`/`UnauthorizedAccessException` or a missing root ⇒ blocked with "Target drive or folder is not reachable." |
| **B2** | `knownPaths.Add` is the first statement in the per-file try again. Test: a known file whose size read throws is skipped once and never re-resolved as unknown. |
| **B3** | `AttributesToSkip = ReparsePoint` only; test enumerates real System/Hidden-flagged files; doc updated. |
| **B4** | Cancel disarms the plan but **keeps the rendered tree**; status and `BlockReason` now say the same true thing ("preview may be stale; press Refresh"). |
| **C1** | `ArgumentException` removed from the skippable filter; blank paths skipped with an explicit `continue`; `IsWithin`/exists checks back outside the try. |
| **C2** | `TaskCanceledException` is only swallowed when the token is *not* cancelled (genuine HTTP timeout); user cancellation propagates as OCE. Both branches tested. |
| **C3** | Cleanup gate reads `plan.IsMove` and the `DeleteEmptySourceFolders` value **snapshotted at Start** (no plan re-stamp, so the confirm-time reference guard isn't tripped by the checkbox). |
| **C4** | Local copies of `_resolveCts` at every cancel site; field nulled before the local is disposed; the UI-thread/SynchronizationContext assumption is now stated in a comment. |
| **D1** | Real ACL test: denies `(OI)(CI)(RX)` via `icacls`, proves the denial bites, asserts the partial preview, removes the ACE in `finally` (also on the elevated-session skip path). |
| **D2** | `ReadCompleted` is `internal`; sidecar journal entries kept (Restore needs journals written today to carry them). |

Also fixed while verifying: the first wave had flipped seven files LF→CRLF — restored, so the diff shows the real change again.

Full test project green, UI Release build clean. GUI smoke still owed before merge (drive-root + UNC source, tab-open lazy init, multi-GB copy responsiveness).

PR #523, opened 2026-08-22
Feature/metadata sync overhaul b thumbnails

Little-God1983:


Little-God1983 (commented):
## Review — Plan B thumbnail pipeline

Read every production hunk plus surrounding context, built `DiffusionNexus.UI` (clean, 0 warnings) and ran the relevant test slices — **828 tests** across `ModelRepositoryTests | Sync | ModelTile | Viewer`, all green.

The shape of this is good: one ladder, one codec, one writer, one retry policy, and `ModelTileViewModel` reduced to a caller instead of a second implementation. The XML docs carry the reasoning rather than restating the code, and several of the sharper edges (the deferred sentinel vs. corrupt, `VideoNoPoster` being soft on purpose, ranking-vs-filtering in the candidate query) are already argued in place.

Five findings below, none of them blocking the design — 4 correctness/robustness, 1 efficiency. Two of them (`IsVideoLike`, the corrupt verdict) touch decisions the doc comments show were deliberate, so they're framed as consequences to weigh rather than mistakes.

PR #525, opened 2026-08-24
feat(sorter): unknown files get the identity chain's header and filename rungs

Little-God1983:
Follow-up to #524. Small and self-contained.

## What this fixes

`SorterMetadataResolver` answers for files that have **no DB row at all** — the "browse any folder" case — via sidecar → per-hash cache → Civitai by-hash. When all three came up empty it gave up and the file sorted into `Unknown\`, even when its own safetensors header names the architecture it was trained on.

#524 landed exactly the two rungs that answer for those files. This wires them in.

```
1. .civitai.info sidecar
2. per-hash disk cache
3. Civitai by-hash API  (+ /models/{id} for tags)
4. safetensors header   ← new (SafetensorsHeaderReader → BaseModelHeaderMap)
5. filename guess       ← new (FilenameBaseModelHeuristic)
```

## A correction to the recorded plan

The follow-up was recorded as *"delete `SorterMetadataResolver`'s ~370-line private chain and route the Sorter through the shared identity chain."* That doesn't hold. `IdentifyModelStep` is entirely DB-row-driven — `SelectAsync(SyncScope…)`, items are `SyncItem(int ModelId, …)`, and it writes to tracked `ModelVersion` entities. There is no path→identity entry point anywhere in the Service, so the resolver answers a question the chain doesn't ask and can't be deleted.

The two rungs are public statics in `DiffusionNexus.Service`, which the UI already references, so they're called directly. The gap was the deliverable, not the duplication.

Also worth noting: the Sorter's **DB-known** branch reads `f.Version.BaseModelRaw` straight from the database, so every model the identity chain filled in on merge already sorts into a real base-model folder with no code change at all. Only the unknown-file path was missing the rungs.

## Design decisions

**Precedence.** The new rungs run only after every authoritative source has come up empty, so a sidecar or Civitai answer is never overruled by a guess about it. The header goes before the file name because it read the actual weights. Two of the six tests are precedence guards and passed before the change — they're there to keep it that way.

**The guess is never cached.** The per-hash cache means *"what Civitai said for this hash"*. The API-failure path deliberately writes no entry so the next pass retries; caching a guess there would kill that retry permanently and freeze the guess as though it were an answer. Re-deriving costs one size-capped (16 MB) header read. `AGuessIsNotCachedSoALaterCivitaiHitStillWins` pins this: offline pass → header answers `SD 1.5`, cache dir is asserted not to exist, next pass → Civitai's `Pony` wins.

**Folder names stay consistent.** Both tables emit verbatim `CivitaiBaseModelCatalog` display labels (reflection tests on both enforce it), so a header-identified file lands in the *same* `Pony\` as a Civitai-identified one, not a folder only it uses.

**Logging** is `Debug`, not `Info` — the resolve loop already logs an `Info` heartbeat every 25 files, and a per-file line at `Info` would flood a 3000-file pass. Each line names which rung answered.

## Sidecar reading left duplicated, on purpose

`SorterMetadataResolver.TryReadSidecar` plucks three fields from a `.civitai.info`; `SidecarMetadataApplier` applies multi-format sidecars to DB entities. Sharing them means a new type with one caller on each side — motion, not de-duplication.

## Tests

Six new, TDD. Four verified RED before the implementation (`FilenameGuessIsTheLastResort…`, `HeaderIdentifiesAFileCivitaiDoesNotKnow`, `TheHeaderBeatsTheFilename…`, `AGuessIsNotCached…`); the two precedence guards passed already, as they should. They reuse `SafetensorsFixture` — the shared byte layout extracted in #524 — rather than re-deriving it.

**4462 passed, 0 failed, 2 skipped** (the opt-in online canaries). UI project builds with 0 warnings.

## Smoke worth doing

Point the Sorter at a folder of LoRAs that aren't in the library and have no `.civitai.info` — self-trained ones especially. They should preview into real base-model folders instead of piling into `Unknown\`. Category still reads Unknown for them, correctly: neither a header nor a file name says anything about category.


Little-God1983:
Followed the "1 of 90" up against the real data rather than guessing, and it was a fair complaint. `bce163f5` expands the table.

## What the numbers actually were

I copied the live library DB and measured. On 328 unidentified files the old table identified **6**. It knew nine families — illustrious, noobai, pony, sdxl, sd15/21/3/35, flux — and that library is mostly LTX, Wan and Qwen. LTX alone was 64 of the misses.

Token frequency across the unidentified names: `ltx` 38, `ltx2` 14, `ltx23` 12, `wan2` 10, `qwen` 10.

With LTXV/LTXV2/LTXV 2.3, Wan Video, Qwen, Hunyuan Video, Chroma, HiDream, Flux.1 Kontext, Flux.2 D and Flux.2 Klein 9B/4B: **6 → 104**.

## This reverses the A3 ruling on `wan*`

A3 dropped every `wan*` key because "Wan Video" can't round-trip through `BaseModelType` and stores as `Other`. On the evidence that reasoning was backwards: a Civitai-identified Wan model **already** stores that exact raw label and that exact `Other`, so withholding it bought nothing and cost the sorter — which files by the raw string — a correct folder.

The gap belongs to the enum, and it is far bigger than those keys: **65 of 86** catalog labels have no member, and ~2,100 of 2,367 versions in that library sit at `Other`. Filed as #526.

The bare `wan` token stays out — `obi_wan_kenobi`. The bare `ltx` key is safe because `latex` tokenizes to `latex`, not `ltx`; both are pinned by test.

## Caveat worth stating plainly

**20 of the 104 are not LoRAs.** VAEs, CLIP encoders, ControlNets and upscalers swept into the library — `LTX23_video_vae_bf16`, `clip_g_hidream`, `Wan2_2_VAE_bf16`, `4x-UltraSharp.pth`. Genuine LoRA wins: **78**.

Naming them changes which folder they land in, not whether they get sorted — they were always being sorted, into `Unknown\`. Filed as #527.

## Follow-ups filed

- **#526** — `BaseModelType` is 65 labels behind; ~2,100 versions unfilterable. Includes the persistence hazard: `Other` is last with implicit value 24, so new members must be appended after it.
- **#527** — non-LoRA support models in the LoRA library.
- **#528** — label each preview folder with its asset kinds and whether it is fully identified.

18 new test rows taken from real filenames. **4491 passed, 0 failed, 2 skipped.**

Little-God1983:
The label idea is in — `6b92f5d1`, closing #528. Kept on this branch rather than a new one since #525 isn't merged and it's the same feature area.

```
Unknown    [LoRA] [VAE] [ControlNet]   ✗
Qwen       [LoRA]                      ✓
LTXV 2.3   [LoRA] [VAE]                ✓
```

Both parts aggregate through the **whole subtree**, not just direct children: a folder's chips are the union of everything beneath it, and the mark ANDs, so a base-model folder is finished only when every category folder under it is. File nodes carry their own kind and mark too — so expanding an unfinished folder shows *which* files are the problem, which was the actual complaint about "224 unidentified" being an opaque number.

"Identified" is read from the candidate **as planned**, so it agrees with the folders the tree is drawing: with *Sort by name* on, a LoRA named into a real folder counts as finished.

## The classifier

`SorterAssetKindClassifier` names the kind from the file name alone. It is fallible by construction, so **nothing it decides moves a file** — it drives the label only. Whether a VAE should be sorted differently, or discovered at all, is #527.

Markers come from names in the real library rather than invented, and the bar for each is that it cannot plausibly occur in a LoRA's own name:

- `clip` counts only as the **first** token — `clip_g_hidream` leads with it, `hair_clip_v1` does not
- a bare `.pth` is **not** an upscaler — `Chris.pth` in that library is an ordinary model
- a leading scale factor **is** — `4x-UltraSharp`, `4xLSDIRplus` — but `4x4` is not, because its remainder is pure digits

Validated against all 328 real filenames: **15 VAEs, 12 text encoders, 5 upscalers, 2 ControlNets, no false positives.**

A test row for `4xLSDIRplus.pth` caught that my first pass missed a scale factor glued straight onto the model name with no separator — it only handled `4x-UltraSharp`, where the factor is its own token.

## Verification

25 new tests. The subtree rollup verified **RED** by dropping the ancestor `Absorb` call — three label tests fail without it.

**4522 passed, 0 failed, 2 skipped.** UI builds with 0 warnings.

## Where this branch stands

| commit | what |
|---|---|
| `e9db56f4` | header + filename rungs for unknown files |
| `8c3ffbd8` | review round: only a real answer licenses a guess; DB-known placeholders |
| `0f3f0eef` | *Sort by name* opt-in with a "will fix 4 of 5" count |
| `bce163f5` | filename table learns LTX / Wan / Qwen / Flux.2 / Kontext / HiDream |
| `6b92f5d1` | folder labels — kind chips + ✓/✗ |

Open follow-ups: #526 (the enum, the big one), #527 (support models in the library).

Colour note for the smoke: ✓ is `#6FCF7F`, ✗ is `#FF6B6B` — the same red the panel already uses for block reasons. If that reads as "error" rather than "unfinished" next to a perfectly fine folder, say so and I'll take it to amber or grey.

DiffusionNexus.UI/Services/Lora/Sorting/SorterMetadataResolver.cs
```diff
@@ -146,20 +318,31 @@ public async Task<ResolvedLoraMetadata> ResolveAsync(string filePath, Cancellati
...
                 $"Civitai by-hash lookup failed for {filePath}: {ex.Message}");
-            return new ResolvedLoraMetadata(null, null, sha);
+
+            // CouldNotAsk. CivitaiClient.GetAsync returns null ONLY for a 404; a rate limit that
+            // survived its three retries, an outage, a non-transient 4xx/5xx and a response-shape
+            // change all arrive here instead, and used to be indistinguishable from "Civitai does
+            // not know this file". This path is serial and unpaced, so one 429 tends to mean every
+            // file after it — and a guess acted on by a move is not undone by "the next pass
+            // retries", because in move mode the file has already left the source folder.
+            return (new ResolvedLoraMetadata(null, null, sha), AuthorityVerdict.CouldNotAsk);
```
Little-God1983: **A failed refresh throws away a base model the cache already had — and the header rung can no longer recover it.**

`CouldNotAsk` is the right verdict for a file nothing has ever answered for, but this branch is also reached for files that *do* have an answer on record. `TryReadCache` (line 286) deliberately declines to serve an entry whose tags are unresolved (`tagsResolved = entry.Tags is not null`), and such entries carry a perfectly good `BaseModel` — they are written whenever `TryReadModelTagsAsync` fails, plus every entry written before tag caching existed.

So the flow is: cache hit with `baseModel: "Illustrious", tags: null` → not served → re-query for the tags → `HttpRequestException` → this line returns `new ResolvedLoraMetadata(null, null, sha)`, discarding `"Illustrious"`, and now *also* returns `CouldNotAsk`, which stops `ResolveAsync` from ever reaching the header rung that used to recover it.

Concretely: Civitai is rate-limiting (the exact scenario this commit is about — "one 429 tends to mean every file after it"), the user previews a library whose tag lookups failed on an earlier pass, and every one of those files sorts into `Unknown\`. In move mode they are physically relocated out of correct folders. Before these commits the same outage at least left the header rung to put them back, so this is a case the fix made strictly worse.

The cached metadata should survive the failed refresh — return `cached! with { Sha256 = sha }` (verdict from `VerdictFor` on its base model) instead of a blank record. The tag list stays unresolved either way, so the next pass still retries it.

DiffusionNexus.Service/Services/Sync/Identity/FilenameBaseModelHeuristic.cs
```diff
@@ -85,18 +85,52 @@ private static readonly (string Prefix, string Label)[] RefinementDecorationPref
...
+        ["ltxv"] = "LTXV",
+        ["ltx"] = "LTXV",
+
+        // Wan. Every version-qualified spelling collapses to the family label: the finer catalog
+        // entries encode t2v/i2v and parameter count, which a file name reports too unreliably to
+        // act on — and a wrong "Wan Video 2.2 I2V-A14B" folder is worse than a right "Wan Video".
+        ["wan25"] = "Wan Video",
+        ["wan22"] = "Wan Video",
+        ["wan21"] = "Wan Video",
+        ["wan2"] = "Wan Video",
```
Little-God1983: **The adjacent-pair candidate set defeats the "no bare `wan` key" guard this table documents.**

The comment above says the bare `wan` token "is NOT here and must not be added: Star Wars character LoRAs (`obi_wan_kenobi`) are extremely common", and `Guess_DoesNotMatchWordsThatMerelyLookLikeAFamily` pins `obi_wan_kenobi` / `wan_kenobi_portrait`. But `Guess()` does not match tokens only — it matches every token **and every adjacent pair and triple**. So `wan2` becomes a candidate for any name where `wan` is followed by a digit token:

- `obi_wan_2.safetensors` → tokens `obi|wan|2` → pair `wan2` → **"Wan Video"**
- `Obi Wan 2.5 - portrait.safetensors` → triple `wan25` → **"Wan Video"**
- `padme_and_wan_21.safetensors` → pair `wan21` → **"Wan Video"**

A `_2` / `_2.5` / `_21` version suffix on a character LoRA is at least as common as the character names the guard protects against, so the guard as written does not hold. And this rung is not behind the sorter's new opt-in on the DB side: `IdentifyModelStep.cs:202` calls the same `Guess()` and writes the result to `ModelVersion.BaseModelRaw` silently.

If the version-qualified spellings are worth keeping, they need to be matched as *tokens* (`wan21`, `wan2.2` → `wan22`) rather than synthesized by the pair/triple set. At minimum add `obi_wan_2` / `obi_wan_2.5` rows to the negative theory — the rows that are there today cannot fail.

DiffusionNexus.Service/Services/Sync/Identity/FilenameBaseModelHeuristic.cs
```diff
@@ -85,18 +85,52 @@ private static readonly (string Prefix, string Label)[] RefinementDecorationPref
...
+        ["chroma"] = "Chroma",
+        ["hidream"] = "HiDream",
+
+        // Flux beyond 1.D/1.S. "kontext" and "klein" are distinctive enough to stand alone;
+        // a bare "klein" is not, because 4B and 9B are different base models and the file name is
+        // the only thing that says which — so only the size-qualified pairs map.
+        ["kontext"] = "Flux.1 Kontext",
+        ["klein9b"] = "Flux.2 Klein 9B",
+        ["klein4b"] = "Flux.2 Klein 4B",
+        ["flux2"] = "Flux.2 D",
```
Little-God1983: **`flux2` as a pair key turns a `_2` version suffix on a Flux.1 LoRA into "Flux.2 D".**

`flux2` is matched against the pair/triple candidate set, and rung 4 runs *before* the `("flux", "Flux.1 D")` prefix rung below. So the second version of an ordinary Flux.1 LoRA changes family:

- `portrait_flux_2.safetensors` → tokens `portrait|flux|2` → pair `flux2` → **"Flux.2 D"** (was "Flux.1 D")
- `mystyle_flux_2_final.safetensors` → same

`name_flux_2` is a very ordinary way to spell "v2 of my Flux LoRA", and it now beats the rung that used to get it right. Same silent-DB-write exposure as the wan keys (`IdentifyModelStep.cs:202`), so this is not gated by the sorter's opt-in.

Worth restricting `flux2` to an exact token (`flux2`, `flux2d` — the `Flux2-Klein-9B` fixture tokenizes to `flux2` on its own) rather than letting the pair set synthesize it, plus a `..._flux_2.safetensors → "Flux.1 D"` test row.

DiffusionNexus.UI/ViewModels/SortPreviewNodeViewModel.cs
```diff
@@ -27,4 +28,51 @@ public partial class SortPreviewNodeViewModel : ObservableObject
...
+    /// <summary>
+    /// False when anything under this node has no base model — the ✗ state. Aggregates by AND
+    /// through the whole subtree, so a base-model folder is only "finished" when every category
+    /// folder beneath it is.
+    /// </summary>
+    [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(StatusTooltip))]
+    private bool _isFullyIdentified = true;
+
+    public string StatusTooltip => IsFullyIdentified
```
Little-God1983: **With "Sort by name" on, the ✓ tells the user a filename guess "has a base model".**

`isIdentified` is read from `move.Candidate.BaseModelRaw` *after* `ApplyNameGuesses` folded the guess in (`LoraSorterViewModel.cs:1280`), so with the opt-in ticked a file whose only evidence is `il_mio_stile` → Illustrious renders identically to one Civitai confirmed — under the tooltip **"Every file here has a base model."**

That is the one screen where a user could check what the lowest-confidence rung is about to do to their files, and it now asserts the opposite. The confirm dialog does not mention it either ("42 files will be moved into X"). `NameGuessHint` in the far-left options rail is the only trace, and it is not attached to the rows it describes.

Round 1's provenance finding was deferred as a product decision, which was fair — but 6b92f5d1 did not leave it neutral, it added a mark that positively claims *identified*. A third state (a `[guessed]` chip, or keeping guessed rows at ✗ with a "named, not read" tooltip) keeps the tree honest, and `SortCandidate.NameGuess` already carries everything needed to tell the two apart at this point.

DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs
```diff
@@ -836,6 +893,44 @@ private async Task<CandidateResolution> ResolveCandidatesAsync(string sourceFold
...
+        var unidentified = 0;
+        var fixable = 0;
+        foreach (var candidate in candidates)
+        {
+            if (!SorterPathBuilder.IsPlaceholderBaseModel(candidate.BaseModelRaw)) continue;
+            unidentified++;
+            if (candidate.NameGuess is not null) fixable++;
+        }
+
+        NameGuessHint = fixable == 0
```
Little-God1983: **`NameGuessHint` survives folder deselection and a failed preview, so it advertises another folder's numbers.**

`UpdateNameGuessHint` is the only writer and it runs only on a preview pass's success path. Two paths clear everything else and leave it behind:

- `RecomputePreviewAsync`'s no-source branch (~lines 424-444) clears `PreviewRoots`, `PreviewSummary`, `DiskSummary`, `HasEnoughSpace`, `BlockReason`, `StatusMessage` and `TransferCount` — but not `NameGuessHint`.
- `DisarmPlan` on the failure path clears the tree and the summaries; the hint stays.

So after deselecting the source, or after "Preview failed: …", the options rail still reads *"5 LoRAs could not be identified — sorting by name will fix 4 of them."* over an empty tree, with numbers belonging to a folder that is no longer selected. Since the whole point of the hint is "what this option is worth on THIS library", a stale pair of numbers is the one thing it must not show. Same stale-state trap `DisarmPlan`'s own doc comment describes for `_lastPlan`; clearing it from `DisarmPlan` plus the deselect branch covers both.

DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs
```diff
@@ -738,8 +775,27 @@ private async Task<CandidateResolution> ResolveCandidatesAsync(string sourceFold
...
+                // and IdentifyModelStep only clears that when a library sync actually runs — it is
+                // due-gated on 30-day windows and attempt caps. Until then the row says nothing, so
+                // ask the file the same two questions the unknown-file branch asks. Without this the
+                // very same LoRA would be identified from its header while sitting in an
+                // unregistered folder and dumped into Unknown\ the moment it is registered, which
+                // is backwards. Costs one size-capped header read per placeholder row; a row that
+                // already carries a base model pays nothing.
+                var baseModelRaw = f.Version.BaseModelRaw;
+                string? nameGuess = null;
+                if (SorterPathBuilder.IsPlaceholderBaseModel(baseModelRaw))
```
Little-God1983: **The new per-row header read lands in the one loop that reports no progress and no heartbeat.**

This is the right fix for the `"???"` rows, but it puts real I/O — a `FileStream` open plus up to `SafetensorsHeaderReader.MaxHeaderBytes` (16 MB) of reads — into the DB-known loop, which unlike the unknown-file loop below it has neither a `progress?.Report(...)` call nor the `ResolveLogInterval` heartbeat.

`ModelFileSyncService` stamps *every* locally-discovered model `BaseModelRaw = "???"`, so on a registered root this is not a rare row — it can be most of the library, i.e. thousands of serial header opens on what may well be a NAS. The overlay sits on the static "Computing preview…" for the whole phase, and the log has nothing between `Resolving candidates under X: N known, M unknown` and the end of the pass. That is exactly the "tell slow from hung" property the heartbeat in the loop below exists for, and the reason it was added there. It is also re-derived from scratch on every source switch and every Refresh.

A `progress?.Report($"Reading headers {n}/{placeholders}…")` plus the same 50-row Info line would cost nothing and close the gap.

DiffusionNexus.Service/Services/Sync/Identity/FilenameBaseModelHeuristic.cs
```diff
@@ -85,18 +85,52 @@ private static readonly (string Prefix, string Label)[] RefinementDecorationPref
...
+        // Wan. Every version-qualified spelling collapses to the family label: the finer catalog
+        // entries encode t2v/i2v and parameter count, which a file name reports too unreliably to
+        // act on — and a wrong "Wan Video 2.2 I2V-A14B" folder is worse than a right "Wan Video".
+        ["wan25"] = "Wan Video",
+        ["wan22"] = "Wan Video",
+        ["wan21"] = "Wan Video",
+        ["wan2"] = "Wan Video",
+
+        ["qwen"] = "Qwen",
+        ["hunyuan"] = "Hunyuan Video",
```
Little-God1983: **`hunyuan` → "Hunyuan Video" mislabels HunyuanDiT, which is a different catalog family rather than a finer version.**

The Wan keys collapse `wan21`/`wan22`/`wan25` to the family label and the justification holds: those catalog entries differ only by version and variant. `hunyuan` is not that case. `CivitaiBaseModelCatalog` carries both `"Hunyuan 1"` (HunyuanDiT, an image model) and `"Hunyuan Video"`, and they are different architectures — so a bare `hunyuan_style_v3.safetensors` now files into `Hunyuan Video\`, which by this PR's own standard is "a wrong folder [that] is worse than Unknown".

Either require a video-qualified spelling (`hunyuanvideo` — the pair set already produces it from `hunyuan_video`), or leave the bare key out. There is no test row for `hunyuan` in either direction today.

DiffusionNexus.UI/Doc/LoraViewer.md
```diff
@@ -878,17 +878,40 @@ If `free < required + 1 GB` safety margin, the available-space bar turns red and
...
+
+The panel says what the option is worth on *this* library before it is turned on — `NameGuessHint`, e.g. *"5 LoRAs could not be identified — sorting by name will fix 4 of them."*, rewording to the past tense once it is on. Both counts are taken from the candidates as resolved, before any guess is folded in, so ticking the box changes the sentence and not the numbers. Because the guess travels with the candidate instead of being baked into it, toggling re-plans off the candidate cache and touches no disk at all.
+
+**DB-known rows get the same two rungs.** `ModelFileSyncService` stamps every locally-discovered model `BaseModelRaw = "???"`, and `IdentifyModelStep` only clears that when a library sync actually runs (it is due-gated on 30-day windows and attempt caps). `ResolveCandidatesAsync` therefore calls `GuessBaseModelFromFileAsync` for any DB-known candidate whose base model is still a placeholder — without it, pointing the sorter at a *registered* LoRA root (which takes the DB-known branch for nearly every file) would do almost nothing for exactly the self-trained LoRAs this exists for.
```
Little-God1983: **This names a method the same commit deleted.**

`GuessBaseModelFromFileAsync` was renamed to `IdentifyFromFileAsync` in 8c3ffbd8 — the commit that added this paragraph. `ResolveCandidatesAsync` calls `_metadataResolver.IdentifyFromFileAsync`, and the split into `FileIdentity.FromHeader` / `FileIdentity.FromName` is the whole point of the rename. The stale name also hides the behaviour that matters here: the DB-known branch applies only the header outright and merely *offers* the name, which is what makes it consistent with the opt-in described two paragraphs above.

DiffusionNexus.UI/Services/Lora/Sorting/SorterMetadataResolver.cs
```diff
@@ -91,14 +112,149 @@ public void ResetPerPassCaches()
...
+    /// </summary>
+    /// <remarks>
+    /// Asked only once every authoritative source has come up empty — and only when they came up
+    /// empty with an <i>answer</i> rather than a failure (see <see cref="AuthorityVerdict"/>) — so a
+    /// sidecar or Civitai answer is never overruled by a guess about it. Public because the sorter's
+    /// DB-known branch needs the same two rungs for rows still carrying the <c>"???"</c> placeholder
+    /// <c>ModelFileSyncService</c> stamps on locally-discovered models; that branch already has the
+    /// database's answer and must not re-hash or re-call the API to get these.
+    /// </remarks>
+    /// <remarks>
```
Little-God1983: **Two `<remarks>` blocks on one member — the second is dropped by every doc consumer.**

`IdentifyFromFileAsync` now carries the new `<remarks>` (lines 196-203) *and* the pre-existing one (204-213) explaining why the answer is deliberately never written to the per-hash cache. XML doc allows one `<remarks>` per member: IDE quick-info and any generated docs render the first and silently drop the second, so the load-bearing "a cached guess would kill that retry permanently and freeze the guess as though it were an answer" rationale disappears at the call site — in a file where these comments are the design record. Merging them into one block is enough; the `<para>` split for the second half is already there.

DiffusionNexus.Tests/Sync/Service/Identity/FilenameBaseModelHeuristicTests.cs
```diff
@@ -74,6 +80,49 @@ public void Guess_StripsKnownModelExtensionsBeforeMatching(string fileName, stri
     /// <c>BaseModelHeaderMapTests.Map_EveryOutputIsACatalogLabel</c> so a future catalog edit can't
     /// silently orphan a heuristic label either.
     /// </summary>
+    /// <summary>
+    /// Real names taken from a live library where the pre-expansion table identified 6 of 328
```
Little-God1983: **The new theory inherited `Guess_EveryOutputIsACatalogLabel`'s doc comment.**

The `<summary>` directly above ("Reflects over `CivitaiBaseModelCatalog`'s private bundled snapshot … asserts every label the heuristic can ever return is a real Civitai display label") belongs to `Guess_EveryOutputIsACatalogLabel` further down. Inserting this theory between them re-parented it, so two `<summary>` blocks now sit on one member: `Guess_IdentifiesTheFamiliesARealLibraryActuallyContains` claims to do catalog reflection, and the test that actually does it — the one guarding exactly the new labels this commit adds — is left undocumented. Moving the block back down restores both.

DiffusionNexus.UI/Services/Lora/Sorting/SorterAssetKind.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        return false;
+    }
+
+    /// <summary>
+    /// A digit run followed by 'x', then either nothing ("4x", from "4x-UltraSharp") or a remainder
+    /// containing a letter ("4xlsdirplus"). The letter is what separates an upscaler from a
+    /// dimension: "4x4" is a remainder of pure digits and must not match.
+    /// </summary>
+    private static bool LeadsWithAScaleFactor(string token)
```
Little-God1983: **`LeadsWithAScaleFactor` matches `2x2x2` — the same dimension shape `4x4` is documented to reject.**

The remainder test is "contains an ASCII letter", and `x` is a letter, so a chained dimension satisfies it: `2x2x2` → digits `2`, then `x`, remainder `2x2` contains `x` → Upscaler. `4x4` correctly returns false; `4x4x4` does not. Nothing moves off this — it drives a chip only — but the rule the doc comment states ("the letter is what separates an upscaler from a dimension: `4x4` is a remainder of pure digits and must not match") is not the rule the code implements, and the `4x4` reasoning in the commit message cannot catch it.

Requiring a letter other than `x`, or rejecting a remainder that is itself a digit/`x` chain, restores the stated rule.

DiffusionNexus.UI/Services/Lora/Sorting/SorterAssetKind.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        "t5", "t5xxl", "umt5", "mistral", "llava", "vl", "textencoder",
+    };
+
+    private static readonly string[] ControlNetTokens =
+    {
+        "controlnet", "ipadapter", "instantx", "redux",
+    };
+
+    private static readonly string[] UpscalerTokens =
```
Little-God1983: **`upscale`, `vl` and `redux` sit below the stated bar for a marker, and have no negative test rows.**

The class remarks set the bar at "it cannot plausibly occur in a LoRA's own name", and `clip`-as-first-token and the bare-`.pth` carve-out are carefully built to meet it. Three markers do not:

- `upscale` (this list) — `hires_upscale_helper.safetensors`, `detail_upscale_v2.safetensors` are ordinary LoRA names, and the longer `upscaler` already covers the real fixture (`ltx-2.3-spatial-upscaler-x2-1.1`).
- `vl` (line 41) — a two-character token, the same length class as the `il` → Illustrious token this PR judged too dangerous to apply without an opt-in. `anime_vl_style` reads as a LoRA; the `qwen_2.5_vl_7b` fixture is already covered by nothing else, so it would need a companion marker rather than the bare pair.
- `redux` (line 45) — Flux Redux *LoRAs* exist, and Redux / IP-Adapter are not ControlNets to begin with, so this one mislabels twice over.

`Classify_DefaultsToLora` has no row for any of the three, so a regression here is invisible. Since the chip is what tells the user "this folder is about to receive something that is not a LoRA", a false chip is the failure mode that matters most.

DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs
```diff
@@ -27,7 +27,10 @@ public partial class LoraSorterViewModel : BusyViewModelBase
     /// <summary>Heartbeat cadence for the unknown-file resolution loop's progress log line.</summary>
     private const int ResolveLogInterval = 50;
 
-    private static readonly string[] ModelExtensions = [".safetensors", ".ckpt", ".pt", ".pth"];
+    // ".sft" is the standard short alias for a safetensors container — SafetensorsHeaderReader
+    // reads it, FilenameBaseModelHeuristic strips it, and ImageResourceHasher enumerates it; only
+    // this list left it out, so a .sft file in a browsed folder was invisible to the whole sorter.
+    private static readonly string[] ModelExtensions = [".safetensors", ".ckpt", ".pt", ".pth", ".sft"];
```
Little-God1983: **Adding `.sft` here leaves a fourth, still-divergent copy of the model-extension list.**

The comment says "only this list left it out", but the four lists still do not agree:

- `FilenameBaseModelHeuristic.KnownModelExtensions` — `.safetensors .pt .ckpt .bin .sft .gguf`
- `BaseModelHeaderMap` (line 31) — the same six
- `SafetensorsHeaderReader.SafetensorsExtensions` — `.safetensors .sft`
- this one — `.safetensors .ckpt .pt .pth .sft`

So `.gguf` and `.bin` are still invisible to the sorter for precisely the reason `.sft` was, and `.pth` is unique to this copy. The next divergence gets found the same way this one did: a user's file silently not appearing. One shared list (with `.pth` folded in, and the safetensors-container subset derived from it) makes the class of bug unrepresentable rather than fixing one instance of it.

DiffusionNexus.UI/Services/Lora/Sorting/SorterMetadataResolver.cs
```diff
@@ -91,14 +112,149 @@ public void ResetPerPassCaches()
...
+
+    /// <summary>
+    /// Whether an authoritative source actually supplied a base model. Uses
+    /// <see cref="SorterPathBuilder.IsPlaceholderBaseModel"/> — the same predicate
+    /// <c>BuildTargetDirectory</c> uses to pick the Unknown bucket — so "resolved enough to skip the
+    /// file's own rungs" and "resolved enough to earn a real folder" cannot drift apart. A "???"
+    /// arriving from a sidecar or an older cache entry is a placeholder to both, not just to one.
+    /// </summary>
+    private static AuthorityVerdict VerdictFor(string? baseModelRaw)
+        => SorterPathBuilder.IsPlaceholderBaseModel(baseModelRaw)
```
Little-God1983: **The predicate this commit consolidated still has two more copies.**

Routing the gate through `SorterPathBuilder.IsPlaceholderBaseModel` is the right call — "resolved enough to skip the file's own rungs" and "resolved enough to earn a real folder" must not be able to drift. But the same rule is spelled out verbatim twice more:

- `DiffusionNexus.Service/Services/Sync/SyncStateDeriver.cs:94` — `IsPlaceholder`, used by `BaseModelWriter.CanFill` and `SidecarMetadataApplier`; i.e. the predicate deciding whether `IdentifyModelStep` may write the very label the sorter later reads.
- `DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs:1098` — `IsPlaceholderBaseModel`.

Three copies of `string.IsNullOrWhiteSpace(x) || x == "???"`, one of them on the other side of the write/read boundary this PR now spans. Since the argument in this very doc comment is that the two ends of one decision have to share a predicate, this is the moment to fold them into one — `SyncStateDeriver.IsPlaceholder` is already documented as the statement of the rule.

Little-God1983:
Round 3: all 14 verified, all fixed at `d8f6bc10`. 4544 tests pass (was 4522), UI builds clean, every new guard RED-verified by perturbation.

Two of them I took to the reference library rather than settling on plausibility, and the data changed the fix:

**The wan guard.** Confirmed exactly as described — `obi_wan_2` → pair `wan2` → Wan Video, and the negative rows that were there could never have failed. But the suggested remedy (match version-qualified spellings as tokens) would have cost far more than it saved: **34 of the 36** wan files in the library spell the version apart — `WAN-2.2-I2V-…`, `wan_22_high_noise`, `Pornmaster_wan 2.2_14b_…` — and all of them are legitimately Wan Video. Token-only matching would drop them and keep exactly one file, a VAE. So the pair synthesis stays and the collision gets named: `NamesObiWan` suppresses every wan key when the tokens say Obi-Wan or Kenobi. Your three rows are in, and they now actually fail without the guard.

**flux2.** Here the suggestion was right, but token-only alone would have regressed a real file: `flux-2-klein-base-9b-fp8.safetensors` tokenizes to `flux|2|klein|base|9b`, so `klein9b` never forms and it would have fallen through to **Flux.1 D** — the wrong family, worse than the coarse answer it has today. Fixed both ends: `flux2` requires the glued spelling, and two new keys `kleinbase9b` / `kleinbase4b` match that file precisely — `Flux.2 Klein 9B-base` is its own catalog label, and **10 versions** in the library carry one. `portrait_flux_2` now correctly reads as Flux.1 D.

**hunyuan** — taken as written. Both labels are in the catalog and they are different architectures. It costs one real file (`hunyuan_20s_horror_900` → Unknown), which is the trade this PR's own standard asks for.

**The ✓ over a guess** is the finding I'd have been most annoyed to ship. Three states now — ✓ read/confirmed, `~` guessed (amber), ✗ no base model — rolled up worst-first through the subtree. It's carried as `SortCandidate.BaseModelIsGuess` rather than by comparing `BaseModelRaw` to `NameGuess`, because a header and a name can agree and that comparison would then report a *read* file as guessed; there's a test pinning that case.

**The cached-answer loss** is the one with the widest blast radius, and your read of it was exact: tag-less entries are common (every pre-tag-caching entry, plus every failed tag lookup), and `CouldNotAsk` was blocking the recovery path as well as discarding the value. The recorded answer now survives with `VerdictFor` applied to it, so a recorded *blank* still licenses the file's own rungs.

**On `vl`** — dropped as a bare marker but kept next to `qwen`. `qwen_2.5_vl_7b_fp8_scaled` is the only file that needs it and `qwen`/`vl` aren't adjacent, so a pair rule wouldn't have reached it. `upscale` (zero files) and `redux` (your double-error argument) are gone outright, all with negative rows.

Both consolidations landed: `ModelFileExtensions` (the `.sft` patch really had left `.bin` and `.gguf` invisible for the same reason) and `SyncStateDeriver.IsPlaceholder` as the one predicate across the write/read boundary.

Still yours: the GUI smoke and the merge. Worth a look — the new `~` is `#E2B14C`, and ✗ stays `#FF6B6B`, the same red the panel uses for block reasons.

DiffusionNexus.Service/Services/Sync/Identity/ModelFileExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+/// that one list left <c>.bin</c> and <c>.gguf</c> invisible for exactly the same reason, and
+/// <c>.pth</c> known to the sorter alone. A shared list makes the whole class of bug
+/// unrepresentable rather than fixing one instance of it: adding a format is one edit here.
+/// </remarks>
+public static class ModelFileExtensions
+{
+    /// <summary>Every recognized model-file extension, lowercase, dot-prefixed.</summary>
+    public static readonly string[] All =
+    {
+        ".safetensors", ".pt", ".pth", ".ckpt", ".bin", ".sft", ".gguf",
```
Little-God1983: **There were seven of these, not four — and the one that feeds the database still omits `.sft`.**

The four folded in here are the identity chain's three plus the sorter's scan. Three more survive verbatim, and the type's own promise ("adding a format is one edit here") is not true while they do:

- `ModelFileSyncService.cs:27` — `[".safetensors", ".pt", ".ckpt", ".pth"]`. This is the scan that *discovers models into the database*, one layer above the sorter. A `.sft` file is still invisible to it, for exactly the reason it was invisible to the sorter — so the file the sorter can now see never gets a row to be seen through. `.bin` and `.gguf` too.
- `StaticFileTypes.ModelExtensions` (`Helper/StaticFileTypes.cs:12`) — `.ckpt`, `.safetensors`, `.pth`, `.pt`; read by `ModelDiscoveryService` (3 call sites) and `JsonInfoFileReaderService` (2).
- `ImageResourceHasher.cs:27` — has `.gguf`, `.bin`, `.sft`, but is the only list missing `.pth`.

The commit message reasons that a shared list makes the class of bug unrepresentable rather than fixing one instance. That holds only once the remaining three are the same list; until then this is a fifth spelling of the same thing that happens to be shared by four call sites.

Separately, worth deciding before folding the rest in: `All` now answers two different questions. For `FilenameBaseModelHeuristic` and `BaseModelHeaderMap` it means *strip this from a name hint* — harmless, and wider is always safer. For `LoraSorterViewModel` it means *enumerate and physically move this file*. Widening the first now silently widens the second: add `.onnx` or `.pkl` here for hint-stripping and the sorter starts relocating those files. `SafetensorsContainers` already shows the shape that avoids it — a list per meaning, not per format.

DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs
```diff
@@ -27,7 +28,10 @@ public partial class LoraSorterViewModel : BusyViewModelBase
     /// <summary>Heartbeat cadence for the unknown-file resolution loop's progress log line.</summary>
     private const int ResolveLogInterval = 50;
 
-    private static readonly string[] ModelExtensions = [".safetensors", ".ckpt", ".pt", ".pth"];
+    // The shared list, not a fourth private copy. Patching this one to add ".sft" left ".bin" and
+    // ".gguf" invisible to the sorter for precisely the reason ".sft" had been — see
+    // ModelFileExtensions.
+    private static readonly string[] ModelExtensions = ModelFileExtensions.All;
```
Little-God1983: **This widens what the sorter moves, and nothing tells the user.**

The finding was that the scan list diverged; the fix also adds `.bin` and `.gguf` to it. Everything this enumeration yields becomes a `SortCandidate` and is planned into a move — so a LoRA root holding `pytorch_model.bin`, a downloaded HF repo folder, or a quantized `.gguf` UNet now has those files relocated into base-model folders, where before they were left alone. `SorterAssetKindClassifier` is name-based, so `pytorch_model.bin` carries a `[LoRA]` chip in the preview; the one signal that would flag it as not-a-LoRA doesn't fire.

That may well be what you want — the same argument that says a `.sft` LoRA should be sortable says a `.gguf` one should be. But it is a behaviour change to a *move* operation arriving inside a bullet about deduplicating constants, and it is the kind that is only noticed after the files have moved.

`Doc/LoraViewer.md:120` still documents the old set ("Scans folders for .safetensors / .pt / .ckpt / .pth files") — one of the two doc lines this commit didn't catch.

Little-God1983 (commented):
Round 3 audit: all 14 findings from round 2 are genuinely fixed, and the two that mattered most — the `CouldNotAsk` cache discard and the preview's ✓ over a guess — are fixed at the right layer (the on-record answer keeps its own verdict; `BaseModelIsGuess` travels on the candidate rather than being re-derived by comparing strings). The new guards have RED-verifiable rows, including the ones that document the reversals (`obi_wan_2`, `portrait_flux_2`, `2x2x2`, `anime_vl_style`).

Two things the consolidation left behind, both the same class of bug it was meant to make unrepresentable. Neither is a regression — they were true before this commit — but the commit is the moment they became checkable.

One residual, no action implied: making `flux2` glued-only settles `portrait_flux_2` at the cost of the mirror case — `flux_2_dev` / `flux.2-dev` now reach the `flux` prefix rung and come out `Flux.1 D`, a confident wrong *family* written to `ModelVersion.BaseModelRaw`. `Guess_ReadsAVersionSuffixAsAVersionNotAFamily` shows this is deliberate and backed by the reference library, so it is only worth revisiting if a separator-spelled Flux.2 file ever turns up.

Little-God1983:
Both true, both fixed at `9f12bdc4`. The second one is a regression I introduced in `d8f6bc10`, and it's the more serious of the two.

**The merged list was the wrong shape.** You put your finger on why: `All` was answering two questions, and a set wide enough for the first is dangerous for the second. Recognizing a *name* as a model's is advisory — over-recognizing costs nothing. Enumerating a *folder* decides which files get physically relocated. So `ModelFileExtensions` now splits:

- **`Sortable`** — enumerate / discover / **move**. `.safetensors .sft .ckpt .pt .pth`
- **`Recognized`** — name hints and resource hashing. Adds `.bin .gguf`
- **`SafetensorsContainers`** — unchanged

It moves to Domain, since discovery, sorting, hashing and the identity chain all read it, and that's where `LocalPathRoots` went for the same reason.

All seven now route through it:

| list | set | delta |
|---|---|---|
| `ModelFileSyncService` | Sortable | gains `.sft` |
| `StaticFileTypes` (5 call sites) | Sortable | gains `.sft` |
| `LoraSorterViewModel` | Sortable | **loses `.bin`/`.gguf`** |
| `ImageResourceHasher` | Recognized | gains `.pth` |
| `FilenameBaseModelHeuristic` | Recognized | — |
| `BaseModelHeaderMap` | Recognized | — |
| `SafetensorsHeaderReader` | Containers | — |

**On `.sft` in discovery** — I'd deliberately left that alone in round 1 as "a bigger behavioural change than the sorter's own enumeration", but you're right that it stops being defensible once the sorter can see one: the sorter would file a model the Viewer could never show. Discovery and the sorter now read the same set. It is a real widening of what the library discovers, and I'm calling it out as such rather than burying it — `.sft` is the short spelling of the container the header reader has always read, so refusing to discover one was inconsistent, not conservative.

Seven tests: six on the invariants (`Sortable` ⊆ `Recognized`, containers ⊆ `Sortable`, `.bin`/`.gguf` recognized but never sortable) and one at the sorter itself asserting a `pytorch_model.bin` and a `.gguf` are not planned for a move while a `.sft` is. RED-verified by pointing the scan back at the wide set — only that test fails.

Doc updated: the discovery line at :120 and a new paragraph on why the split exists, so the next person widening a format knows which of the two they're widening.

**On the `flux_2_dev` residual** — agreed, and thanks for flagging it as non-actionable. The library has zero separator-spelled Flux.2 files and three `flux`+digit names that are training-step suffixes, so the asymmetry currently favours the glued rule. If a `flux_2_*` shows up it'll want the same treatment `hunyuan_video` got — a qualified pair rather than a bare family key.

4554 tests pass.

DiffusionNexus.Domain/Utilities/ModelFileExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <summary>
+    /// Files the application will enumerate, discover into the library, and MOVE. Deliberately
+    /// narrower than <see cref="Recognized"/>: adding an entry here relocates users' files.
+    /// </summary>
+    /// <remarks>
+    /// <c>.sft</c> is here because it is the same container as <c>.safetensors</c> under its short
+    /// name — the header reader has always read it, so a library that could not discover one was
+    /// inconsistent with its own identity chain, not conservative.
+    /// </remarks>
+    public static readonly string[] Sortable =
```
Little-God1983: **`.sft` is now a first-class citizen everywhere except the two places that enumerate `*.safetensors` by hand.** Both predate this commit; the widening is what makes them reachable.

- `DuplicateScanner.cs:39` — `Directory.EnumerateFiles(folder, "*.safetensors", …)`. This one is **live**: a duplicated pair saved as `.sft` is never reported, while the sorter will now discover, plan and file both copies. The sorter's own dedup (`NormalizeHash` / `SameContent`) and the Duplicate scanner now disagree about what a model file is.
- `JsonInfoFileReaderService.cs:33` — `f.Extension == ".safetensors" || f.Extension == ".pt"` picks the group's primary weights file, and anything else sets `NoMetaData = true` and skips the metadata fetch. Latent rather than live: its only caller is `DuplicateScanner`, which has already filtered to `.safetensors` — but it has been silently dropping `.ckpt`- and `.pth`-only models the whole time, and `.sft` just joined them. The grouping two methods away (line 97/103) reads the shared `Sortable` set, so the two halves of the same file now disagree.

Neither is in this diff and neither needs to be in this PR. Flagging because the commit message's claim — discovery and the sorter read the same set, so a file the sorter would file can never be one the library refuses to discover — holds at that boundary and stops one call site short of the Duplicate scanner.

Separately, small: `Sortable`, `Recognized` and `SafetensorsContainers` are `public static readonly string[]`, so the reference is fixed but the contents are not — `ModelFileExtensions.Sortable[0] = ".exe"` compiles, from any of the seven consumers. For a type whose entire job is being the one source of truth this is worth `ImmutableArray<string>` or `FrozenSet<string>` (the latter also makes `Matches` a lookup instead of a scan). Same argument the split itself makes: prefer the shape where the mistake cannot be written.

Little-God1983 (commented):
Both round-3 findings are fixed, and the split is the right shape — `Sortable` vs `Recognized` is the distinction the flat list was papering over, the sorter is back to the narrow set with a test that pins the regression, and `Domain/Utilities` is the correct home (dependency direction is right, `ImplicitUsings` is on there so the bare `Path` resolves, and the `LocalPathRoots` cref lands in the same namespace). The invariant tests are the good kind: `SortableIsASubsetOfRecognized` and `WeightFormatsWeDoNotMoveAreRecognizedButNotSortable` fail on the exact edit that would re-introduce the bug, rather than restating the arrays.

One comment below, on the edges the widening leaves — both pre-existing, one live — and a small note on the arrays themselves.

On the count: `StaticFileTypes.GeneralExtensions` and `SupportedTypes.ModelTypesByPriority` are the two remaining lists that touch model extensions and are correctly *not* routed through this — the first answers a third question (associated files, unioned with `ModelExtensions` at `ModelDiscoveryService.cs:118`), and the second has no consumers at all. Worth deleting `ModelTypesByPriority` while the area is open: it is dead, and it is the one list that still says `.onnx` and `.bin` are model types.

Issue #521, opened 2026-08-21
Metadata sync & Civitai download pipeline overhaul (one clean refactor, safe for existing users)


Little-God1983:
## TL;DR

"Download Metadata" re-processes the whole library on every run because the sync **cannot distinguish "we already have this" from "we never tried"**. Measured on the real library today (2,577 models, log `log-20260821.txt`, run 11:14–11:35):

| | Selected | Actually useful | Cost |
|---|---|---|---|
| Phase 1 metadata by hash | 3 | 1 | 2 erroring files retried forever |
| Phase 1b sidecar re-process | 91 | 1 | 90 have no sidecar, re-read every run |
| Phase 3 tag backfill | 68 | 6 | 62 have zero tags on Civitai → asked again every run (68 × 1.5 s pacing ≈ 100 s of nothing) |
| Phase 4 thumbnails | **1,046** | ~0 | **19 of the 21 minutes**; **2 GB of preview videos re-downloaded** (largest 331 MB) to regenerate 440 thumbnails that were already in the DB; 497 × `The 'file' scheme is not supported` stack traces |

The second run, minutes later, selected exactly the same 68 / 1,046 again.

This issue is the plan for **one clean refactor** of the metadata/thumbnail sync *and* the three Civitai download paths that feed it, designed to be **safe for existing users** (additive schema only, no mass re-sync after upgrade, no overwrites, downgrade-tolerant). Target outcome: first run after upgrade does only the genuinely missing work; every following run is *"Nothing to do"* in seconds; bulk sync never downloads a video again.

---

### RC1 — Phase 4 asks the *screen*, not the database
[...]

PR #529, opened 2026-08-25
Metadata sync overhaul Plan D: one Civitai download path (WP5)

Little-God1983:
## What this is

Plan D of the #521 metadata-sync overhaul — **spec WP5: one Civitai download path**. Every download path (Download LoRA dialog, Browse queue, Detail panel, waitlist→queue, Pipelines installer) now goes through a single `ICivitaiModelDownloader` with one primary-file pick, one collision policy, one persister, one SHA256 verification, a Tags+Thumbnails completion sync, and an `ILibraryChangeNotifier` signal that finally makes the Installed tab update after queue downloads.

Plan doc: `docs/superpowers/plans/2026-08-24-metadata-sync-overhaul-D-download-path.md` · Spec: `docs/superpowers/specs/2026-08-21-metadata-sync-overhaul-design.md` §4.4 / S4 / S5 / D3.

## The 8 tasks (each TDD'd + independently reviewed, all findings fixed)

1. **Shared primitives** — `CivitaiVersionFiles.PickPrimary` (was 8 copies), `FileFormatMapper` (3), `FileSizeFormatter.FormatKilobytes` (3), SHA copies → `FileHasher`
2. **`LoraPathBuilder` → Service** (was `SorterPathBuilder`, gains `includeBaseModel`) + **`DownloadCollisionPolicy`** (the queue's proven algorithm + identical-content reuse); the three picker VMs delegate — downloads and the sorter can no longer disagree about where a file belongs
3. **`ICivitaiApiKeyProvider`** — replaces five `GetApiKeyAsync` copies
4. **S5 in the persister** — `IsUserEdited` guards mirroring `CivitaiMetadataApplier`; re-downloads no longer clobber hand-edited names/tags/base models
5. **`ICivitaiModelDownloader` + `ILibraryChangeNotifier`** — the one path; owns the `IDownloadCoordinator` enqueue (D3), kills four caller-side coordinator/TCS copies
6. **Detail panel + toolbar migrate** — the ~530-line inline downloader clone deleted; category tag `"2000"` no longer becomes a category; viewer rebuilds via a coalesced notifier subscription (one rebuild per batch)
7. **Queue + waitlist + pipeline migrate** — plus the browse-card bug: "enqueue selected versions" was a silent no-op on primary-search cards
8. **Riders** — `.sft`-blind scans (`DuplicateScanner`, `JsonInfoFileReaderService`) routed through `ModelFileExtensions.Sortable`; spec WP1/WP2/WP5 checkboxes; `Doc/LoraViewer.md` corrected; coordinator activity-log guard

## Behavior changes worth knowing (all reviewed + deliberate)

- **S4 everywhere:** dialogs/detail no longer delete-and-overwrite a colliding file — `{stem}_{versionId}` rename, byte-identical files reused without re-downloading (`ReusedExisting`)
- Download folder names are sanitized; blank base model + folder toggle → `Unknown\`; a category literally named "Unknown" makes no segment (sorter parity)
- Every download ends with a poster thumbnail + tags (completion sync) and an Installed-tab refresh — no manual refresh
- Cancelled downloads read as cancelled, not failed (including coordinator-side cancel); hash mismatches say so and keep the file for inspection
- Discovery/duplicate scanning widened to the full `Sortable` set (`.sft`, `.ckpt`, `.pt`, `.pth`)
- GB/KB size display precision unified via the shared formatter (F1→F2 GB, F0→F1 KB)

## Verification

- **4646 passed / 0 failed / 2 skipped** (pre-existing opt-in online canaries), `dotnet test DiffusionNexus.Tests -c Release`; UI/Service/Infrastructure build 0 warnings
- Per-task review gates + final whole-branch review (fix waves included): S4 collision matrix on real files, S5 matrix on real EF/SQLite, downloader orchestration matrix incl. a real Windows ACL-deny regression test, D3 verified by tree-wide grep (exactly one Civitai coordinator enqueue site)

## Still yours (manual smokes before merge)

1. Download LoRA dialog into a folder holding a same-named different file → `_versionId` rename, original untouched; identical file → reused, still registered
2. Browse queue batch → tiles appear in Installed tab with posters, ONE rebuild (not one per job)
3. Detail panel → Download this version → tab flips blue, tile refreshes
4. Pipeline Civitai LoRA install → `.civitai.info` still written, status bar reports
5. Re-download over a user-edited model → name/tags/base model survive

Known follow-ups (non-blocking, from the final review): `ActualSha256` stamp after restart-rehydration should come from the downloader's verified hash; persist-before-verify leaves a DB row registered on HashMismatch; `DownloadTrigger.Waitlist` currently unreachable (waitlist jobs run as BrowseQueue).


DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
```diff
@@ -1275,7 +736,7 @@ private void BuildLocalVersionTabs(ModelTileViewModel tile)
 
         foreach (var version in tile.Versions)
         {
-            // Map local files to CivitaiModelFile so PersistDownloadedModelAsync has data
+            // Map local files to CivitaiModelFile so a download of this version has file data
```
Little-God1983: **Blocker — silent weight destruction.**

`BuildLocalVersionTabs` maps local `ModelFile`s to `CivitaiModelFile` **without `Hashes`**, so every detail-panel download of a local version runs the new collision policy and step-7 verification blind.

Concretely: a local version's file goes missing (`IsLocalFileValid=false` → `tab.IsDownloaded=false`), the user clicks *Download this version*. `request.File = primaryFile` carries `Hashes = null`, so `expectedSha256` is null:

* `DownloadCollisionPolicy.MatchesAsync` returns `false` unconditionally → the plain name is never reused → the target becomes `{stem}_{Version.Id}{ext}`, where `Version.Id` is `version.CivitaiId ?? 0`.
* Two local-only versions in the same folder therefore both resolve to `{stem}_0.safetensors`.
* `suffixedMatches` is false (no hash to match), so `DownloadFileAsync` does `File.Delete(targetPath); File.Move(...)` — **the second download silently overwrites the first model's weights.**
* Step 7 verification is skipped as well, so the S4 guarantee is inert on exactly this path.

This is the same class of bug as the earlier filename-collision incident, reintroduced through a lossy DTO mapping. The local `ModelFile` rows do carry hashes — they need to travel into the synthesised `CivitaiModelFile`.

DiffusionNexus.UI/Services/Download/CivitaiModelDownloader.cs
```diff
@@ -0,0 +1,340 @@
...
+                            $"Step 7: SHA256 verified for {resolution.TargetPath} ({actual})");
+                    }
+                }
+                catch (Exception ex)
+                {
+                    _logger?.Warn(LogCategory.Download, LogSource, $"SHA256 verification failed: {ex.Message}");
+                }
+            }
+
+            if (status == DownloadStatus.HashMismatch)
```
Little-God1983: **Blocker — a corrupt file is registered in the library as valid.**

The DB row is written *before* verification can reject it: `LoraDownloadService.DownloadFileAsync` calls `PersistDownloadedModelAsync` and `completed()` before it returns, so Model/ModelVersion/ModelFile already exist with `IsLocalFileValid = true`.

Step 7 then computes a mismatching SHA256 and this `return Finish(...)` skips steps 8–10 (`ResolveModelIdAsync`, `RunCompletionSyncAsync`, `_notifier.NotifyModelDownloaded`).

Net result on a truncated or tampered transfer: the corrupt model stays permanently registered as a valid library entry — just invisible on the Installed tab until a manual refresh, after which it looks like any other installed LoRA. The verification detects the problem and then leaves the evidence in place. `HashMismatch` needs to roll the row back, or at minimum clear `IsLocalFileValid`, rather than only short-circuiting the happy path.

DiffusionNexus.UI/Services/Download/CivitaiModelDownloader.cs
```diff
@@ -0,0 +1,340 @@
...
+            var reusedPersist = await _downloadService
+                .PersistDownloadedModelAsync(resolution.TargetPath, request.Version, request.ExistingModelId)
+                .ConfigureAwait(false);
+            if (reusedPersist != MetadataPersistOutcome.Complete)
+            {
+                _logger?.Debug(LogCategory.Download, LogSource,
+                    $"Step 5: metadata for the reused file came back {reusedPersist}");
+            }
+
+            status = DownloadStatus.ReusedExisting;
```
Little-God1983: **Reporting gap — `ReusedExisting` claims full success on failed metadata.**

When the file on disk is already byte-identical (`ExistingContentMatches`), a `MetadataPersistOutcome.Failed` — e.g. the Civitai model-page fetch 500s — only reaches a Debug log at 128–131; `status` is still `ReusedExisting` and `DownloadOutcome.Success` is `true`.

The identical condition on the transfer path (~236) yields `CompletedMetadataIncomplete` + *"Done — no metadata"* + a Warn log. So the queue tile shows *"Already downloaded"*, the library gets a row with no description/tags/preview, and nothing warns — which is exactly the silence `CompletedMetadataIncomplete` was introduced to eliminate.

The reuse path also never gets `TryRepairMetadataByHashAsync`, which the transfer path picks up for free inside `DownloadFileAsync`.

DiffusionNexus.UI/Services/CivitaiBrowser/CivitaiDownloadQueue.cs
```diff
@@ -521,14 +521,6 @@ private async Task RunJobAsync(CivitaiDownloadJob job, CancellationToken ct)
             targetDir = Path.Combine(folders[0], string.IsNullOrWhiteSpace(job.BaseModel) ? "Unsorted" : job.BaseModel);
         }
         Directory.CreateDirectory(targetDir);
```
Little-God1983: **Crash — one unwritable destination aborts the entire batch.**

`Directory.CreateDirectory(targetDir)` is outside the `try` that starts on the next line. A destination on a disconnected share or a read-only folder throws `UnauthorizedAccessException`/`IOException` here; `RunGatedAsync` has only a `finally`, so it propagates into `Task.WhenAll(tasks)` in `StartAllAsync`, which has no `try`.

Consequence: `RaiseCountsChanged()`, `Persist()` and the batch summary are all skipped, every remaining job stays `Queued`, and it surfaces as an unhandled relay-command fault.

It is also redundant now — `CivitaiModelDownloader` step 3 (87–100) creates the same directory and converts these exact exceptions into a `Failed` outcome with *"target directory unavailable"*. Deleting this line fixes the crash **and** removes the duplication.

DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
```diff
@@ -444,200 +439,42 @@ private async Task DownloadSelectedVersionAsync()
...
+                    // Keep the panel's own fallback name when the version carries no named file,
+                    // instead of the downloader's synthesized "model_{id}.safetensors".
+                    FileNameOverride = string.IsNullOrWhiteSpace(primaryFile?.Name)
+                        ? $"{ModelName}_{tab.Label}.safetensors"
+                        : null,
+                };
 
-            using (response)
+                var outcome = await _modelDownloader.DownloadAsync(request).ConfigureAwait(false);
+                if (outcome.Success && outcome.FinalPath is not null)
```
Little-God1983: **Regression — the migrated detail-panel download reports nothing on Failed / Cancelled / HashMismatch.**

`if (outcome.Success)` and no `else`. Civitai returns 403 on a gated model → the button just stops spinning: no message, no dialog, no status text.

The inline downloader this replaced called `activityLog?.CompleteDownloadProgress(false, $"Download failed: {...}")` and `taskHandle?.Fail(...)`, and the sibling migration in `LoraViewerViewModel.DownloadLoraAsync` (985–993) maps every `DownloadStatus` to a user-visible `SyncStatus` line. The detail panel got no equivalent, so the typed-outcome plumbing lands here without a consumer.

DiffusionNexus.UI/Services/CivitaiBrowser/CivitaiDownloadQueue.cs
```diff
@@ -549,127 +541,101 @@ private async Task RunJobAsync(CivitaiDownloadJob job, CancellationToken ct)
...
-                catch (Exception ex)
+
+                switch (outcome.Status)
                 {
-                    _logger?.Warn(LogCategory.Download, "CivitaiQueue", $"SHA256 verification failed: {ex.Message}");
-                }
-            }
+                    case DownloadStatus.Completed:
+                        job.ProgressPercent = 100; // the throttled progress adapter's last report may sit below 100
+                        job.ActualSha256 ??= job.ExpectedSha256; // verified inside the downloader
```
Little-God1983: **`ActualSha256` records a digest that was never computed against the bytes on disk.**

The comment at 551–556 says the enqueue-time `ExpectedSha256`/`FileName` snapshot and the downloader's expectation *"agree by construction"* — they don't. After restart-rehydration `civVersion` is re-fetched and `primaryFile = CivitaiVersionFiles.PickPrimary(civVersion)` is a **fresh** pick; only `FileNameOverride` comes from the job snapshot.

If the re-picked file carries no `Hashes.SHA256`, `CivitaiModelDownloader` logs *"no expected SHA256 — verification skipped"* and returns `Completed`. Lines 588/598 then write the stale enqueue-time digest into `ActualSha256` under the comment *"verified inside the downloader"*, and that unverified digest is persisted to the queue file.

Either thread the actually-computed hash back out of the outcome, or leave `ActualSha256` null when the downloader skipped verification.

DiffusionNexus.UI/Services/Download/CivitaiModelDownloader.cs
```diff
@@ -0,0 +1,340 @@
...
+            // A bad or unwritable destination is the caller's problem to report, not an exception
+            // five call sites would each have to remember to catch.
+            _logger?.Warn(LogCategory.Download, LogSource,
+                $"Cannot create target directory {request.TargetDirectory}: {ex.Message}");
+            return Finish(new DownloadOutcome(DownloadStatus.Failed, null, null, false, "target directory unavailable"), fileName);
+        }
+
+        // 4 — collision policy: reuse our own bytes, never overwrite someone else's (S4).
+        var expectedSha256 = Present(file?.Hashes?.SHA256);
+        var resolution = await DownloadCollisionPolicy
```
Little-God1983: **`OperationCanceledException` still escapes `DownloadAsync`, contradicting the design note just above it.**

Step 4's collision resolution is awaited outside a `try`. `DownloadCollisionPolicy.ResolveAsync` → `FileHasher.Sha256UpperAsync(path, ct)` hashes an existing multi-GB file, and `MatchesAsync` catches only `IOException`/`UnauthorizedAccessException` — so cancelling mid-hash throws straight out of this await.

That breaks the promise at 92–94 (*"not an exception five call sites would each have to remember to catch"*). Both migrated call sites are unguarded: `ModelDetailViewModel.DownloadSelectedVersionAsync` (447) and `LoraViewerViewModel.DownloadLoraAsync` (972) wrap the call in `Task.Run` with only a `finally`, so it becomes an unobserved task fault with zero user feedback.

DiffusionNexus.UI/Services/Download/CivitaiModelDownloader.cs
```diff
@@ -0,0 +1,340 @@
...
+    /// download that reached disk has succeeded whatever the follow-up does.
+    /// </summary>
+    private async Task RunCompletionSyncAsync(int modelId, CancellationToken ct)
+    {
+        if (_librarySync is null) return;
+
+        var acquired = false;
+        try
+        {
+            await CompletionGate.WaitAsync(ct).ConfigureAwait(false);
```
Little-God1983: **Efficiency — `CompletionGate` throttles the transfers themselves, not just post-processing.**

Step 9 blocks on the process-wide `CompletionGate` and then runs `PlanAsync` + `ExecuteAsync` (Civitai tag fetch + thumbnail downloads) **while the caller still holds its download slot** — both `CivitaiDownloadQueue._gate` and the coordinator slot.

With `_gate = 2` and a 20-job batch, job N+2's *transfer* cannot start until job N's *network metadata sync* has finished and released the gate. Every download's throughput ends up bounded by the slowest thumbnail fetch ahead of it. The completion sync wants to run outside the transfer slot — after the outcome is returned, or batched.

DiffusionNexus.UI/Services/CivitaiBrowser/CivitaiDownloadQueue.cs
```diff
@@ -549,127 +541,101 @@ private async Task RunJobAsync(CivitaiDownloadJob job, CancellationToken ct)
...
-                // the boolean result that the coordinator wants.
-                return await tcs.Task.ConfigureAwait(false);
-            }
-
-            if (coordinator is not null)
+                File = primaryFile,
+                FileNameOverride = job.FileName,
+                TaskName = $"Download {job.ModelName} ({job.VersionName})",
+            };
+            var progressAdapter = new Progress<DownloadProgress>(p => Dispatcher.UIThread.Post(() =>
```
Little-God1983: **Ordering — the FIFO assumption in the comment below does not hold for `Progress<T>`.**

`progressAdapter` is a `Progress<DownloadProgress>` whose handler itself calls `Dispatcher.UIThread.Post`. `Progress<T>.Report` never invokes the handler inline — it posts to the captured `SynchronizationContext` (or the thread pool). So a progress update is *two* hops (Report → handler → dispatcher post), while the terminal state at 575 is a *single* direct dispatcher post.

A progress report issued microseconds before `DownloadAsync` returns can therefore have its second hop enqueued **after** the terminal post, leaving the tile showing `512.0 / 512.0 MB` at 99% instead of *Done*/*Cancelled*.

Separately, the inner `Dispatcher.UIThread.Post` is a redundant double-marshal whenever the adapter is constructed on the UI thread — which it is for the first two jobs, since `_gate.WaitAsync` completes synchronously when a slot is free.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -892,58 +957,110 @@ private async Task DownloadLoraAsync()
...
-        var taskName = $"Downloading {fileName}";
 
+        // The one download path (spec §4.4) owns the coordinator enqueue — so this must NOT
+        // wrap it in one — plus verification, persistence and the library-changed signal.
+        // The Installed tab rebuilds off that signal, so there is no manual refresh here.
+        var downloader = _modelDownloader;
         _ = Task.Run(async () =>
         {
-            var tcs = new TaskCompletionSource<bool>();
+            var progress = new Progress<DownloadProgress>(p =>
```
Little-God1983: **Same `Progress<T>` double-hop ordering defect as the queue.**

The handler is constructed inside `Task.Run`, so `Report` hops to the thread pool before the handler's dispatcher post is even enqueued, while the terminal `Dispatcher.UIThread.Post` at 987 is a single hop from the awaiting thread. A report issued just before completion can land after it, leaving the toolbar permanently reading *"Downloading foo.safetensors: 512.0 / 512.0 MB"* instead of *"Downloaded foo.safetensors"*.

DiffusionNexus.UI/Services/CivitaiBrowser/CivitaiDownloadQueue.cs
```diff
@@ -549,127 +541,101 @@ private async Task RunJobAsync(CivitaiDownloadJob job, CancellationToken ct)
...
-                // own _gate already throttles to 2. The coordinator's slot is acquired
-                // immediately since we're well under its cap.
-                await coordinator.EnqueueAsync(taskName, RunDownloadAsync, ct).ConfigureAwait(false);
-            }
-            else
+                job.ProgressPercent = p.Percent;
+                job.StatusMessage = p.Message;
+            }));
+            var outcome = await _downloader!.DownloadAsync(request, progressAdapter, ct).ConfigureAwait(false);
+            job.TargetPath = outcome.FinalPath;
```
Little-God1983: **Off-thread write to a bound observable property.**

`DownloadAsync` is awaited with `ConfigureAwait(false)`, so this line runs on a thread-pool thread. `TargetPath` is an `[ObservableProperty]` that drives the bound `DisplayPath`, so this raises `PropertyChanged` for a live Avalonia binding off the UI thread — precisely the hazard the next few lines take care to avoid by marshalling the status/percent/hash writes through the dispatcher. The collision `_logger?.Warn` at 566–570 sits in the same off-thread window.

DiffusionNexus.UI/Services/CivitaiBrowser/CivitaiDownloadQueue.cs
```diff
@@ -521,14 +521,6 @@ private async Task RunJobAsync(CivitaiDownloadJob job, CancellationToken ct)
             targetDir = Path.Combine(folders[0], string.IsNullOrWhiteSpace(job.BaseModel) ? "Unsorted" : job.BaseModel);
```
Little-God1983: **Reuse — this fallback is the one remaining hand-rolled path build in the download stack.**

`Path.Combine(folders[0], job.BaseModel)` takes `job.BaseModel` straight from Civitai with no sanitisation. A base model carrying an invalid filename character (or a placeholder like `"???"`) either throws out of the `Directory.CreateDirectory` below or creates a folder the sorter will never agree with — the sorter routes everything through `LoraPathBuilder.BuildTargetDirectory` with `SanitizeFolderName` and the `Unknown` bucket.

That download-vs-sorter drift is what Task 2 set out to eliminate, so this should be `LoraPathBuilder.BuildTargetDirectory(folders[0], job.BaseModel, job.Category, includeBaseModel: true, includeCategory: false)`.

DiffusionNexus.UI/Services/LoraDownloadService.cs
```diff
@@ -685,25 +722,10 @@ private async Task<bool> TryRepairMetadataByHashAsync(string filePath, int? exis
...
-        {
-            using var scope = App.Services.GetRequiredService<IServiceScopeFactory>().CreateScope();
-            var settingsService = scope.ServiceProvider.GetRequiredService<IAppSettingsService>();
-            return await settingsService.GetCivitaiApiKeyAsync();
-        }
-
-        return _settingsService is not null
-            ? await _settingsService.GetCivitaiApiKeyAsync()
-            : null;
+        _apiKeyProvider ??= new CivitaiApiKeyProvider(App.Services?.GetService<IServiceScopeFactory>(), _settingsService);
```
Little-God1983: **Duplication — five copies of `GetApiKeyAsync` became five copies of the same lazy provider construction.**

`_apiKeyProvider ??= new CivitaiApiKeyProvider(App.Services?.GetService<IServiceScopeFactory>(), _settingsService);` is duplicated verbatim here, in `CivitaiBrowserViewModel.cs:1158` and `DownloadLoraDialogViewModel.cs:346`, with a two-argument variant in `LoraViewerViewModel.cs:1242` and `ModelDetailViewModel.cs:1087`.

`App.axaml.cs:950` already registers `ICivitaiApiKeyProvider` as a singleton, so each of these could resolve it (`App.Services?.GetService<ICivitaiApiKeyProvider>() ?? new CivitaiApiKeyProvider(...)`) from one shared helper. As written, the next change to the fallback rules — a cache, a timeout — still has five sites to find, which is the duplication this task set out to kill.

DiffusionNexus.Service/Services/Lora/LoraPathBuilder.cs
```diff
@@ -1,14 +1,14 @@
...
 /// <summary>
 /// Pure path construction for the LoRA Sorter: folder naming, sanitization
 /// (nothing in the download path sanitizes — this is deliberately new), and the
 /// deterministic collision rename convention shared with
-/// CivitaiDownloadQueue.ResolveCollisionFreeTargetPathAsync
+/// DownloadCollisionPolicy
 /// ({stem}_{versionId}{ext}), so re-runs are idempotent.
 /// </summary>
-public static class SorterPathBuilder
+public static class LoraPathBuilder
```
Little-God1983: **Repo convention — cross-project rename with no `[Obsolete]` forwarder.**

`.github/copilot-instructions.md`: *"When a refactoring task is issued then mark the method/function/class you are refactoring as obsolete and to be removed later to avoid code duplication/dead code."*

The public `DiffusionNexus.UI.Services.Lora.Sorting.SorterPathBuilder` was deleted outright and re-created here as `DiffusionNexus.Service.Services.Lora.LoraPathBuilder` — namespace *and* assembly change — forcing 9 call sites across `LoraSortPlanner`, `SorterMetadataResolver`, `LoraSorterViewModel` and three download view models into the same commit. The same applies to the removed `CivitaiDownloadQueue.ResolveCollisionFreeTargetPathAsync` and the changed `CivitaiDownloadQueue(LoraDownloadService?)` ctor signature.

A thin `[Obsolete]` shim forwarding to `LoraPathBuilder` would have satisfied the rule and let the call-site migration land as its own change.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -386,6 +441,16 @@ public LoraViewerViewModel(
...
         _scopeFactory = scopeFactory ?? App.Services?.GetService<IServiceScopeFactory>();
+        _apiKeyProvider = apiKeyProvider ?? App.Services?.GetService<ICivitaiApiKeyProvider>();
+        _modelDownloader = modelDownloader ?? App.Services?.GetService<ICivitaiModelDownloader>();
+        _changeNotifier = changeNotifier ?? App.Services?.GetService<ILibraryChangeNotifier>();
+
+        // Every download path raises this after persisting, so the Installed tab no longer
+        // depends on the downloading surface remembering to tell it.
+        if (_changeNotifier is not null)
+        {
+            _changeNotifier.ModelDownloaded += OnLibraryModelDownloaded;
```
Little-God1983: **Lifetime — subscription without teardown.**

`ILibraryChangeNotifier` is registered `AddSingleton` while `LoraViewerViewModel` is `AddScoped`, and this class has no `Dispose` and no detach anywhere.

Harmless today, because only one instance is resolved (`App.axaml.cs:1224`, root scope) — but the registration lifetime does not enforce that. Any second resolution (a new DI scope, a re-created viewer, a future navigation model) permanently roots the old view model through the singleton's invocation list and doubles every coalesced rebuild, with each stale instance still hitting the database on every download. The class already implements this teardown pattern for `DetailViewModel`'s events.

Little-God1983 (commented):
## Code review — Plan D (one Civitai download path)

The consolidation itself reads well: a single typed `DownloadOutcome`, one collision policy, one path builder. The findings below are mostly about the seams where the old paths were removed — several of the guarantees this PR introduces turn out to be inert or contradicted on at least one surviving route.

**Two I'd hold the merge on**, both inside the guarantees the PR exists to provide:

1. `ModelDetailViewModel.BuildLocalVersionTabs` drops `Hashes`, so detail-panel downloads of a local version run the collision policy *and* verification blind — two local-only versions collapse onto `{stem}_0.safetensors` and the second download deletes the first model's weights.
2. On `HashMismatch` the DB row has already been written with `IsLocalFileValid = true` inside `DownloadFileAsync`, so a corrupt file stays registered as a valid library entry.

Then: an unwritable destination taking down the whole batch (and the `CreateDirectory` that causes it is now redundant), `ReusedExisting` reporting success over failed metadata, the detail panel reporting nothing on any failure, `ActualSha256` persisting a digest that was never computed, cancellation escaping `DownloadAsync` into two unguarded `Task.Run`s, `CompletionGate` throttling transfers rather than post-processing, two `Progress<T>` ordering defects and one off-thread observable write — plus reuse items (queue fallback bypassing `LoraPathBuilder`, five copies of the API-key provider construction, the `SorterPathBuilder` rename without an `[Obsolete]` shim) and a missing notifier unsubscribe.

15 inline comments follow. Nothing was changed on the branch.

<sub>Reviewed with Claude Code.</sub>

Little-God1983:
All 15 findings triaged, verified real, and 14 fixed in `9c2570f8..679424bd` (7 commits). Suite **4669 passed / 0 failed / 2 skipped**; UI + Service still 0 warnings. Per finding:

**Blockers**
- **Lossy `Hashes` mapping (3852836705)** — the synthesized `CivitaiModelFile` in `BuildLocalVersionTabs` now carries all four local hashes (single mapping helper, tested). The deeper half was also real and got a policy fix: for `versionId <= 0` (local-only versions map `CivitaiId ?? 0`), the `{stem}_0` suffix is meaningless, so `DownloadCollisionPolicy` now walks `{stem}_2, _3, …` and returns only a hash-proven match or the **first free name** — it can no longer hand back an existing unproven path on that route. `versionId > 0` semantics are byte-identical to before.
- **Corrupt file registered valid (3852836712)** — `HashMismatch` now marks the persisted `ModelFile` row(s) at the final path `IsLocalFileValid = false` + `LocalFileVerifiedAt`, in its own scope, with a real-SQLite test. The row keeps its metadata for repair; the file no longer poses as installed.

**Behavior / correctness**
- **`ReusedExisting` hiding failed metadata (3852836718)** — reuse now returns `ReusedExisting` only on `Complete`; `Partial`/`Failed` → `CompletedMetadataIncomplete` + Warn, so the tile says "Done — no metadata" exactly like the transfer path. One accepted gap remains, deliberately: the reuse path still doesn't get `TryRepairMetadataByHashAsync` (it's private to the transport's failure flow); the Installed tab's Download Metadata covers it.
- **Batch-killing `CreateDirectory` (3852836725)** — deleted; the downloader's own step 3 already converts those exceptions to a `Failed` outcome. New batch-survival test.
- **Silent detail panel (3852836733)** — non-success outcomes now map to `StatusMessage` (cancelled / hash-mismatch / failed with the outcome's error), mirroring the toolbar's switch.
- **Unverified `ActualSha256` (3852836737)** — `DownloadOutcome` gained `VerifiedSha256`: the computed digest on a verified transfer, the proven expectation on a content-proven reuse, null when verification was skipped. The queue stamps only from it; the "agree by construction" comment is gone. Your read was right — after rehydration the pick is fresh and only `FileNameOverride` was snapshot.
- **OCE escaping step 4 (3852836741)** — the whole body from the collision probe on is now guarded by `catch (OperationCanceledException) when (ct.IsCancellationRequested)` → `Cancelled` outcome. The filter deliberately does NOT swallow an OCE when the token wasn't signalled — that's a bug, not an outcome.
- **CompletionGate throttling transfers (3852836747)** — the Tags+Thumbnails sync + notifier now run detached (`Task.Run`, `CancellationToken.None` — a finished download's metadata shouldn't die with the job token); `DownloadAsync` returns right after the ModelId resolve, releasing the queue gate. Tests await an internal `LastCompletionTask` seam. One named consequence: the detail panel's immediate refresh can now render before the poster thumbnail lands; the notifier-driven Installed-tab rebuild picks it up.
- **`Progress<T>` double-hop, both sites (3852836756, 3852836766)** — replaced with a single-hop `UiThreadProgress<T>` (`Report` posts directly to the dispatcher), so progress and terminal posts share one FIFO queue; the queue handler additionally drops reports once the job left `Downloading`. Your double-marshal observation held too — gone with the adapter.
- **Off-thread `TargetPath` write (3852836774)** — moved inside the terminal dispatcher post. The collision Warn stays where it was (logging is thread-safe; no binding involved).
- **Hand-rolled fallback path (3852836780)** — routed through `LoraPathBuilder.BuildTargetDirectory(..., includeCategory: false)`. Named delta: blank/`"???"` base model now lands in `Unknown\` (was `Unsorted\`) and names are sanitized — sorter parity, which is the point.
- **Five lazy provider constructions (3852836788)** — one `CivitaiApiKeys.Resolve` helper that prefers the registered singleton; all five sites route through it. Future fallback-rule changes now have one home.
- **Notifier without teardown (3852836804)** — `LoraViewerViewModel` implements `IDisposable`, detaching from the stored notifier (idempotent, tested); DI scope disposal drives it. The "single instance today" argument stops being load-bearing.

**The one pushback — `[Obsolete]` forwarder (3852836793).** The convention's goal is no dangling duplicates/dead code during refactors. All 9 call sites migrated inside this same PR and nothing outside this solution consumes `SorterPathBuilder`, the queue's collision method, or the old ctor — so a forwarder would be born dead, which the same instruction argues against. If you want the shim anyway (e.g. to keep the rename reviewable as its own commit series, or for out-of-tree tooling), say so and I'll add `[Obsolete]` forwarders for all three in one commit.

One honesty note: a single test failed once in the first full-suite run after the wave and passed in two consecutive identical re-runs; the runner's summary truncated the name before I could capture it. If a flake shows up in CI, that's the trail.

Still yours: the GUI smokes from the PR description, and the merge.

PR #530, opened 2026-08-25
Plan E (WP6): sync plan dialog, report view, settings

Little-God1983:
## What this is

Plan E of the metadata-sync overhaul (#521) — spec WP6, the last work package before WP7 acceptance. "Download Metadata" now shows the user what a sync will do **before** it runs, reports faithfully **after** it, and the retry windows + thumbnail parallelism are user-tunable.

Design: `docs/superpowers/specs/2026-08-21-metadata-sync-overhaul-design.md` §4.6 (decision D4: dialog, not auto-start). Plan doc: `docs/superpowers/plans/2026-08-25-metadata-sync-overhaul-E-sync-ui.md`.

## What shipped

- **SyncPlanDialog** gates the bulk sync: discovery pre-run (so counts are truthful) → plan → dialog with one checkbox row per step (count, description, estimate), a *Force re-check…* expander that re-plans live, and the all-zero state showing "Library is up to date — nothing to do" with the last-run stamp and no Start.
- **SyncReportDialog** after every started run: per-step Planned/Processed/Succeeded/Skipped/Failed table + collapsed expanders of failures grouped by step with per-item reasons — never a bare "N failed". Cancelled runs show as partial. The discovery count is folded back into the run's report so the status line, summary, and dialog all agree.
- **Settings → LoRA Viewer → Metadata Sync**: not-identified retry window (7/14/30/60/90 d, default 30), error retry window (1/3/7 d, default 1), thumbnail download parallelism (1–8, default 4). Values round-trip through settings export/import; `LastLibrarySyncAt` is machine-local and deliberately excluded from both export and the save whitelist.
- **Thumbnails-step bounded parallelism** (deferred here from Plan B): `Parallel.ForEachAsync`, clamp 1–8, Thumbnails **only** — a dedicated test pins that paced Civitai API steps can never parallelize.
- The scroll-time thumbnail gate and the per-tile button now honor the settings-derived retry policy; `SyncPlan.HasWork` fixed (was constant-true); `UploadThumbnailAsync` routed through `ThumbnailWriter` (last path around the one-writer; a user upload now clears a stale failure verdict); 4 additive `AppSettings` columns + migration + self-heal entries; `Doc/LoraViewer.md` updated.

## ⚠ Pre-existing bug found (NOT fixed here — deliberately out of scope)

**Every save from the Settings page nulls `AppSettings.LastBackupAt` in the DB.** The Settings VM never carries the field and `AppSettingsService.SaveSettingsAsync` blind-copies the null onto the tracked entity (`AppSettingsService.cs:305`). Verified real on develop; predates this branch. The new `LastLibrarySyncAt` column is protected from this bug class two ways (VM capture/replay + whitelist omission). Wants its own bugfix PR.

## Notes from the final whole-branch review (no action needed to merge)

- The downloader's post-download completion sync still builds `SyncOptions` without a `RetryPolicy` (uses `Default`) — inert in practice (a fresh model is due under any window); consistency follow-up.
- A rare race can surface "A library sync is already running." as an error in the detail panel (pre-existing WP2 behavior, caught safely).
- Old settings-export files (pre this PR) import the three new settings as their defaults — unavoidable for newly added fields.

## Verification

- 4,725 passed / 0 failed / 2 skipped (opt-in online canaries); UI + IntegrationTests build 0 warnings.
- Every task subagent-reviewed; concurrency + dialog VMs adversarially probed (replan serialization, lost-update, thread-affinity, tally integrity under parallelism — empirically, not by reading); load-bearing guards RED-verified by mutation.

## Manual smokes owed before/at merge (spec §6 items 1–3, 7 preview)

1. Press Download Metadata on the reference library → dialog counts look right; Start runs only checked steps; report groups failures.
2. Press again → up-to-date dialog with last-run stamp, no Start; a Force from that state produces counts and un-hides Start.
3. Settings changes alter the next run's plan (e.g. concurrency, windows).
4. Cancel mid-run → partial report; next plan resumes the remainder.
5. Visual: plan dialog regrows when a force un-hides rows (SizeToContent), Expander styling, report table alignment.

### DIFF
--- a/DiffusionNexus.Domain/Models/SettingsExportData.cs
+++ b/DiffusionNexus.Domain/Models/SettingsExportData.cs
@@ -159,8 +159,18 @@ public static class SettingsExportSchema
     /// <summary>
     /// Current schema version. Bump when adding new fields.
     /// v2: split AutoBackupEnabled into BackupDatasetImagesEnabled + BackupDatabaseEnabled.
+    /// v3: metadata-sync retry windows + thumbnail concurrency.
     /// </summary>
-    public const int CurrentVersion = 2;
+    /// <remarks>
+    /// Bumping is cheap in both directions. The importer enforces only
+    /// <see cref="MinSupportedVersion"/> — there is no upper bound — so v1 and v2 files still
+    /// import into this build, and an older build reading a v3 file simply ignores the three
+    /// unknown JSON properties. What the bump buys is the distinction a later migration would
+    /// otherwise have no signal for: whether a field is absent because the file predates the
+    /// feature or because the user chose nothing — exactly what
+    /// <see cref="SettingsExportData.LegacyAutoBackupEnabled"/> needed for the v1→v2 split.
+    /// </remarks>
+    public const int CurrentVersion = 3;
 
     /// <summary>
     /// Minimum schema version that can still be imported.
--- a/DiffusionNexus.Domain/Models/SettingsExportData.cs
+++ b/DiffusionNexus.Domain/Models/SettingsExportData.cs
@@ -50,6 +50,21 @@ public sealed record SettingsExportData
     public bool MergeLoraSources { get; init; }
     public int LoraUpdateCheckStalenessDays { get; init; } = 3;
 
+    // ?? Metadata Sync ????????????????????????????????????????
+
+    /// <summary>Days before the bulk metadata sync re-checks a not-identified model. Default 30.</summary>
+    public int SyncNotIdentifiedRetryDays { get; init; } = 30;
+
+    /// <summary>Days before the bulk metadata sync retries a model whose last attempt errored. Default 1.</summary>
+    public int SyncErrorRetryDays { get; init; } = 1;
+
+    /// <summary>Parallel thumbnail downloads during a bulk sync. Default 4.</summary>
+    public int SyncThumbnailConcurrency { get; init; } = 4;
+
+    // Deliberately no LastLibrarySyncAt here: it is machine-local bookkeeping
+    // (stamped by the sync flow itself, not a user-chosen setting), so it is
+    // not part of the portable export/import contract.
+
     // ?? LoRA Sort ????????????????????????????????????????????
 
     public string? LoraSortSourcePath { get; init; }
--- a/DiffusionNexus.Domain/Services/Sync/SyncRetryPolicy.cs
+++ b/DiffusionNexus.Domain/Services/Sync/SyncRetryPolicy.cs
@@ -8,6 +8,17 @@ public sealed record SyncRetryPolicy(TimeSpan NotIdentifiedRetryAfter, TimeSpan 
 {
     public static SyncRetryPolicy Default { get; } = new(TimeSpan.FromDays(30), TimeSpan.FromDays(1), 3);
 
+    /// <summary>
+    /// Builds a policy from the user's saved settings. Floors both windows at one day: zero would
+    /// mean "always due", which turns every plan into a full re-run — when that is wanted it is a
+    /// Force checkbox on the plan dialog, not a saved setting. <see cref="MaxErrorAttempts"/> stays
+    /// at the default: it caps API hammering inside one run window and is not a user-facing knob.
+    /// </summary>
+    public static SyncRetryPolicy FromDays(int notIdentifiedDays, int errorDays) =>
+        new(TimeSpan.FromDays(Math.Max(1, notIdentifiedDays)),
+            TimeSpan.FromDays(Math.Max(1, errorDays)),
+            Default.MaxErrorAttempts);
+
     /// <summary>Whether an identity attempt is due given the stored outcome.</summary>
     public bool IsIdentifyDue(SyncOutcome outcome, DateTimeOffset? checkedAt, int attempts, DateTimeOffset now, bool force)
     {
--- a/DiffusionNexus.Domain/Services/Sync/SyncRetryPolicy.cs
+++ b/DiffusionNexus.Domain/Services/Sync/SyncRetryPolicy.cs
@@ -9,14 +9,29 @@ public sealed record SyncRetryPolicy(TimeSpan NotIdentifiedRetryAfter, TimeSpan 
     public static SyncRetryPolicy Default { get; } = new(TimeSpan.FromDays(30), TimeSpan.FromDays(1), 3);
 
     /// <summary>
-    /// Builds a policy from the user's saved settings. Floors both windows at one day: zero would
-    /// mean "always due", which turns every plan into a full re-run — when that is wanted it is a
-    /// Force checkbox on the plan dialog, not a saved setting. <see cref="MaxErrorAttempts"/> stays
-    /// at the default: it caps API hammering inside one run window and is not a user-facing knob.
+    /// Builds a policy from the user's saved settings. Clamps both windows to 1–3650 days:
+    /// <see cref="MaxErrorAttempts"/> stays at the default, because it caps API hammering inside
+    /// one run window and is not a user-facing knob.
     /// </summary>
+    /// <remarks>
+    /// The floor: zero would mean "always due", which turns every plan into a full re-run — when
+    /// that is wanted it is a Force checkbox on the plan dialog, not a saved setting.
+    /// <para>
+    /// The cap: these values are not guaranteed to come from the settings combo boxes.
+    /// <c>SettingsExportService.ImportAsync</c> copies them straight out of a JSON file with no
+    /// range check, and <c>AppSettingsService.SaveSettingsAsync</c> persists them unvalidated —
+    /// so a hand-edited or corrupted <c>settings.json</c> carrying <c>2147483647</c> reached
+    /// <see cref="TimeSpan.FromDays"/>, which throws above ~10 675 199 days. Every Download
+    /// Metadata press then died with "TimeSpan overflowed because the duration is too long", and
+    /// the tiles' startup read silently fell back to the default: an unusable button with no way
+    /// to see why. Ten years is past every meaningful horizon, so clamping degrades instead of
+    /// throwing — the same "the service owns the sane range" stance already applied to
+    /// <c>ThumbnailConcurrency</c> in <c>LibrarySyncService</c>.
+    /// </para>
+    /// </remarks>
     public static SyncRetryPolicy FromDays(int notIdentifiedDays, int errorDays) =>
-        new(TimeSpan.FromDays(Math.Max(1, notIdentifiedDays)),
-            TimeSpan.FromDays(Math.Max(1, errorDays)),
+        new(TimeSpan.FromDays(Math.Clamp(notIdentifiedDays, 1, 3650)),
+            TimeSpan.FromDays(Math.Clamp(errorDays, 1, 3650)),
             Default.MaxErrorAttempts);
 
     /// <summary>Whether an identity attempt is due given the stored outcome.</summary>
--- a/DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
+++ b/DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
@@ -1018,10 +1018,18 @@ public partial class LoraViewerViewModel : BusyViewModelBase, IDisposable
                 $"User started sync: steps [{string.Join(", ", chosen.Steps)}] forces " +
                 $"[I:{chosen.ForceIdentify} T:{chosen.ForceTags} Im:{chosen.ForceImages} Th:{chosen.ForceThumbnails}]");
 
-            // Re-planned, not the dialog's plan: it is cheap, the dialog may have been open for
-            // minutes, and the ticks and forces the user came back with select a different set of
-            // items than the one the dialog was built from.
-            var runPlan = await Task.Run(() => _librarySync.PlanAsync(SyncScope.Library, chosen, ct), ct);
+            // The dialog's own plan when it can vouch for it, and only otherwise a third pass
+            // (F13). PlanAsync runs SelectAsync on every requested step, and each of those is a
+            // repository query over the whole library plus per-item retry filtering — this button
+            // was paying for it at least twice, and once more for every Force the user toggled,
+            // even though each toggle had already re-planned with the exact options BuildResult
+            // then returned. The dialog withholds its plan whenever its counts no longer match its
+            // toggles (a failed re-plan), and then this plans again.
+            //
+            // A minutes-old plan is safe: RunStepAsync re-selects per step at execution time, so
+            // the plan's counts reach nothing but the report's Planned column.
+            var runPlan = choice.Plan
+                          ?? await Task.Run(() => _librarySync.PlanAsync(SyncScope.Library, chosen, ct), ct);
 
             var progress = new UiProgress<LibrarySyncProgress>(this, p =>
                 SyncStatus = $"{SyncReport.Label(p.Step)} [{p.Index}/{p.Total}] {p.CurrentItem}");
--- a/DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
+++ b/DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
@@ -120,6 +120,20 @@ public partial class LoraViewerViewModel : BusyViewModelBase, IDisposable
     private CancellationTokenSource? _metadataSyncCts;
 
     /// <summary>
+    /// The user's retry windows, read from settings at startup and again at the top of every bulk
+    /// sync. Cached because the tiles' scroll-fetch gate asks per activation — a settings read per
+    /// tile passing the viewport is not a thing to do — and because the per-tile fetch has to judge
+    /// its un-forced steps by the same windows the sync run does.
+    /// </summary>
+    private SyncRetryPolicy _scrollRetryPolicy = SyncRetryPolicy.Default;
+
+    /// <summary>
+    /// The startup read that fills <see cref="_scrollRetryPolicy"/>. Test seam: tests await it
+    /// instead of relying on it having finished.
+    /// </summary>
+    internal Task ScrollRetryPolicyLoad { get; private set; } = Task.CompletedTask;
+
+    /// <summary>
     /// True from the moment this ViewModel starts a run — bulk or per-tile — until it finishes.
     /// </summary>
     /// <remarks>
@@ -494,6 +508,10 @@ public partial class LoraViewerViewModel : BusyViewModelBase, IDisposable
 
         _ = InitializeBaseModelFilterAsync();
         _ = LoadDestinationFoldersAsync(destination);
+
+        // Once, at startup: the tiles' scroll-fetch gate needs the user's retry windows before the
+        // first bulk sync of the session ever happens, and it must not read settings per tile.
+        ScrollRetryPolicyLoad = LoadScrollRetryPolicyAsync();
     }
 
     /// <summary>
@@ -790,7 +808,36 @@ public partial class LoraViewerViewModel : BusyViewModelBase, IDisposable
             ScopeFactory: sp?.GetService<IServiceScopeFactory>(),
             DialogService: TryResolveDialogService(sp),
             Clipboard: sp?.GetService<DiffusionNexus.Installer.SDK.Shared.Services.IClipboardService>(),
-            UiScheduler: sp?.GetService<IUiScheduler>());
+            UiScheduler: sp?.GetService<IUiScheduler>(),
+            // A delegate, not the value: tiles built now must see a policy re-read later (the
+            // startup load landing, or a bulk sync picking up changed settings) without a rebuild.
+            RetryPolicyProvider: () => _scrollRetryPolicy);
+    }
+
+    /// <summary>
+    /// Reads the saved sync settings once at startup so the tiles' scroll-fetch gate judges by the
+    /// user's retry windows from the first scroll, not only after the first bulk sync.
+    /// </summary>
+    /// <remarks>
+    /// Through <see cref="UseSettingsServiceAsync"/> for a fresh scope: this runs concurrently with
+    /// the other two startup tasks, which share the constructor-injected service's single
+    /// non-thread-safe DbContext. Failure is silent and harmless — the built-in default stands in.
+    /// </remarks>
+    private async Task LoadScrollRetryPolicyAsync()
+    {
+        try
+        {
+            var settings = await UseSettingsServiceAsync(s => s.GetSettingsAsync()).ConfigureAwait(false);
+            if (settings is null) return;
+
+            _scrollRetryPolicy = SyncRetryPolicy.FromDays(
+                settings.SyncNotIdentifiedRetryDays, settings.SyncErrorRetryDays);
+        }
+        catch (Exception ex)
+        {
+            _logger?.Debug(LogCategory.General, "LoraViewer",
+                $"Could not read the sync retry settings; using the defaults: {ex.Message}");
+        }
     }
 
     /// <summary>
@@ -806,16 +853,20 @@ public partial class LoraViewerViewModel : BusyViewModelBase, IDisposable
     }
 
     /// <summary>
-    /// Runs a full library metadata sync through <see cref="ILibrarySyncService"/>: plans the
-    /// work (discover new files, identify unknown ones, fetch tags and images), executes the
-    /// plan, then rebuilds the grid once from the database.
+    /// Runs a full library metadata sync through <see cref="ILibrarySyncService"/> as a
+    /// conversation: scan for new files, plan the rest, ask the user what to run, run it, record
+    /// that it happened, rebuild the grid once, and show what came of it.
     /// </summary>
     /// <remarks>
     /// The viewer owns none of the sync logic any more (#521 WP2) — no hashing, no Civitai
     /// calls, no per-tile persistence. Everything the old phases did is a step in the service,
     /// which records per-model state so a second run skips what has already been checked
-    /// instead of re-asking Civitai about the whole library. Thumbnails are not a step yet
-    /// (Plan B); on-screen tiles keep pulling their own preview on activation.
+    /// instead of re-asking Civitai about the whole library.
+    /// <para>
+    /// Discovery is a separate, un-negotiable pre-run (Plan E): a scan cannot be counted before it
+    /// has run, so it is the one step nobody can be asked about — and running it first is what lets
+    /// every count in the dialog include the files it just found.
+    /// </para>
     /// </remarks>
     [RelayCommand(CanExecute = nameof(CanStartMetadataSync))]
     private async Task DownloadMissingMetadataAsync()
@@ -847,7 +898,7 @@ public partial class LoraViewerViewModel : BusyViewModelBase, IDisposable
         {
             IsBusy = true;
             IsCancellable = true;
-            BusyMessage = "Syncing with Civitai...";
+            BusyMessage = "Scanning source folders…";
             SyncStatus = "Planning sync...";
 
             // The first plan after the upgrade also backfills a state row for every model that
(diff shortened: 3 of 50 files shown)


DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+
+            var dialogVm = new SyncPlanDialogViewModel(
+                plan,
+                baseOptions,
+                replanAsync: o => Task.Run(() => _librarySync.PlanAsync(SyncScope.Library, o, ct), ct),
+                settings.LastLibrarySyncAt,
+                discovered,
+                _logger);
+
+            var choice = await dialogService.ShowSyncPlanDialogAsync(dialogVm);
```
Little-God1983: **Cancelling the plan dialog leaves the files discovery just inserted invisible in the grid.**

Discovery now runs *before* the dialog and commits new `Model` rows to the database, but `RebuildTilesFromDatabaseAsync()` is only reached on the run path. Every early return after the scan — the cancel branch below, the `Dialog service not available.` branch, and both `ReportServiceAlreadyRunning` returns — skips it, and nothing else refreshes the grid (`ILibraryChangeNotifier.ModelDownloaded` is raised only by `CivitaiModelDownloader`, never by `DiscoverFilesStep`).

Failure scenario: user drops 12 new LoRAs into a source folder → presses **Download Metadata** → the dialog says "12 new files discovered" → user presses **Cancel** → the 12 models are in the DB but the viewer still shows the old tile set until a manual Refresh or an app restart, and the status line says "Sync cancelled — nothing was run.", which is now false.

Suggested fix: rebuild whenever `discovered > 0`, regardless of which path the flow leaves by (e.g. a `try/finally` around everything after the discovery run, or an explicit `if (discovered > 0) await Task.Run(RebuildTilesFromDatabaseAsync);` in each early-return branch).

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+                discoverReport = await Task.Run(() => _librarySync.ExecuteAsync(discoverPlan, null, ct), ct);
+            }
+            catch (InvalidOperationException ex)
+            {
+                ReportServiceAlreadyRunning(ex);
                 return;
             }
 
-            // Plan B replaces this with a confirmation dialog showing the same numbers.
+            var discovered = discoverReport.NewFilesDiscovered;
```
Little-God1983: **The discovery pre-run's failures are thrown away — only its count survives.**

`discoverReport` carries `Failures`, `UnexpectedFailures` and `Cancelled`, and all of them are dropped here; only `NewFilesDiscovered` is folded back into the run report at line 1019. `DiscoverFilesStep.ExecuteOneAsync` deliberately returns `SyncItemResult.Failure(ex.Message)` for `IOException` / `UnauthorizedAccessException` / `DbUpdateException` / `TaskCanceledException` precisely so a report can show them — before Plan E, discovery was a step of the single run and that row reached `report.Failures`, the status line (`· N failed`) and `DescribeOutcome`.

Failure scenario: one LoRA source folder is on a disconnected network share → `DiscoverNewFilesAsync` throws `IOException` → the step records a failure and `DiscoveredCount` stays 0 → the user sees a plan dialog with no discovered line, no failure anywhere, a report dialog that never mentions the scan, and concludes the library is fully scanned. The only trace is a log entry.

Suggested fix: carry `discoverReport.Failures` / `UnexpectedFailures` into the reconstructed `SyncReport` at line 1019, and surface them when the user cancels at the dialog too.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+            SyncReport discoverReport;
+            try
             {
-                SyncStatus = UpToDateStatus;
-                _logger?.Info(LogCategory.Network, "CivitaiSync", "Sync plan is empty — nothing to do");
+                discoverReport = await Task.Run(() => _librarySync.ExecuteAsync(discoverPlan, null, ct), ct);
+            }
+            catch (InvalidOperationException ex)
+            {
+                ReportServiceAlreadyRunning(ex);
```
Little-God1983: **`catch (InvalidOperationException)` is far wider than the single-flight gate it is meant to catch.**

`ReportServiceAlreadyRunning` is reached by *any* `InvalidOperationException` escaping `PlanAsync`/`ExecuteAsync`, not just the `_gate.Wait(0)` one — and there are several realistic sources: `scope.ServiceProvider.GetRequiredService<IAppSettingsService>()` in `ResolveApiKeyAsync`, `GetRequiredService<IUnitOfWork>()` in every step's `SelectAsync`, and any LINQ `Single()`/`First()` over an empty sequence inside a step.

Failure scenario: a DI registration regression makes `GetRequiredService<IUnitOfWork>()` throw inside `ThumbnailsStep.SelectAsync` → the user is told "A metadata sync is already running.", it is logged at **Info** (not Error, and without the exception object), the button re-enables, and pressing it again reproduces the same wrong message forever. A genuine bug is disguised as a "not now".

Suggested fix: throw a dedicated `SyncAlreadyRunningException` from the gate (or match the sentinel / re-check `_librarySync.IsRunning`) and let every other `InvalidOperationException` fall through to the generic `catch (Exception)` that logs at Error.

DiffusionNexus.UI/ViewModels/SyncPlanDialogViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+    /// <see cref="WhenReplanSettles"/> cover re-plans queued while another is running — which
+    /// means the prologue of <see cref="ReplanAfterAsync"/> before its first await must stay
+    /// free of anything that can re-enter this method, or the inner queue link is silently
+    /// dropped from the chain.
+    /// </summary>
+    private void QueueReplan()
+    {
+        var options = OptionsWith(AllRowKinds);
+        var previous = _replanTask;
+        _replanTask = ReplanAfterAsync(previous, options);
```
Little-God1983: **Start can be pressed with stale counts in the gap between two queued re-plans.**

`IsReplanning = true` is only set *after* `await previous` resumes, and that continuation is posted to the dispatcher rather than run inline. So the sequence for two quick toggles is: T1 sets `IsReplanning = true` → T1's `finally` sets it back to **false** and completes → (dispatcher turn — the UI is live, `CanStart` is true, the rows still show T1's counts) → T2's continuation finally sets `IsReplanning = true` again.

Failure scenario: user ticks *Force tags*, then immediately ticks *Force image records*, then clicks **Start** in that gap. `BuildResult()` filters by `r.IsSelected && r.Count > 0`, and the Images row is still 0 from the T1 plan, so `SyncStepKind.FetchImages` is silently dropped from the run the user just asked for — with the dialog showing numbers that do not match what was ticked.

Suggested fix: raise the flag at queue time (in `QueueReplan`, before it returns) with a pending counter, and lower it only when the counter reaches zero — so `CanStart` stays false for the whole queued chain, not just while one link is executing.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+                return;
+            }
+
+            // The scan was its own run; fold its count back in so every projection of this run
+            // agrees. Not `with`: Summary is a get-only auto-property, and the copy constructor
+            // keeps the stale one — the dialog would print "Discovered 0" above "40 new files
+            // discovered" and the status bar would carry the same contradiction.
+            report = new SyncReport(report.Plan, report.Steps, report.Failures, report.Cancelled,
```
Little-God1983: **The report's elapsed time silently excludes the discovery pass — and is rendered with the *estimate* formatter.**

The discovered *count* is folded back here, but `report.Elapsed` still comes only from the second `ExecuteAsync`'s stopwatch. On a real library the folder scan is often the slowest part of the whole button press, and it is now a separate run whose `Elapsed` is discarded. `SyncReportDialogViewModel` then renders it through `SyncPlanDialogViewModel.FormatDuration`, which prefixes `~` (an estimate marker) and emits a bare unlabelled string.

Failure scenario: a 4 000-file library takes 3 min to scan and 40 s to sync → the user waits ~4 min and the report says `~40 s`, unlabelled, with a tilde implying it is an approximation rather than a measurement.

Suggested fix: add `discoverReport.Elapsed` when rebuilding the `SyncReport`, and give the report dialog its own exact-duration formatter (`Elapsed: 3 min 42 s`) instead of reusing the plan dialog's estimate formatter.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+            // to the catch below would skip the rebuild AND the report dialog, which is the F1
+            // failure documented under the rebuild, re-entered through a new door. A timestamp is
+            // not worth everything the run achieved, so a failed stamp costs only the timestamp.
+            if (!report.Cancelled)
+            {
+                try
+                {
+                    await Task.Run(() => UseSettingsServiceAsync(async s =>
+                    {
+                        await s.UpdateLastLibrarySyncAtAsync(DateTimeOffset.UtcNow, CancellationToken.None)
```
Little-God1983: **`LastLibrarySyncAt` is stamped for *any* non-cancelled run, but the plan dialog presents it as "Last full sync".**

The whole point of Plan E's dialog is that the user chooses a subset of steps — yet the stamp is unconditional on `!report.Cancelled`, and `SyncPlanDialogViewModel.LastRunText` renders it as `Last full sync: …`. That stamp is the only staleness signal the next dialog shows.

Failure scenario: user unticks everything except *Thumbnails* and runs a 20-second thumbnail top-up. Identify/Tags/Images have never run. Next week the dialog opens and states "Last full sync: 2026-08-25 14:03" — the user reasonably concludes the metadata is a week old when in fact it was never fetched at all.

Suggested fix: either stamp only when the run covered all four planned kinds (`chosen.Steps.SetEquals(PlannedStepKinds)`), or relabel the dialog line to "Last sync" and record which steps it covered.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+                settings.LastLibrarySyncAt,
+                discovered,
+                _logger);
+
+            var choice = await dialogService.ShowSyncPlanDialogAsync(dialogVm);
+            if (!choice.Confirmed || choice.Options is null)
+            {
+                // The button did do something — it scanned — so it owes an answer either way. Which
+                // answer depends on whether there was anything to decline in the first place.
+                SyncStatus = plan.HasWork ? "Sync cancelled — nothing was run." : UpToDateStatus;
```
Little-God1983: **The cancel-at-dialog status is computed from the *pre-force* plan, so it can contradict what the user just saw.**

`plan` is the plan built before the dialog opened. The dialog's Force expander re-plans live and can turn an all-zero plan into one with thousands of items — but `plan` is never updated with that result.

Failure scenario: library is up to date → dialog shows "Library is up to date — nothing to do" → user expands *Force re-check…* and ticks **Thumbnails**, the dialog re-plans and shows `40 Thumbnails`, Start un-hides → user changes their mind and presses **Cancel** → the status line reads "Library is up to date — nothing to do", flatly contradicting the 40 the dialog was showing a second earlier.

Suggested fix: base the wording on the dialog VM's current state (`dialogVm.IsUpToDate`) rather than on the stale `plan`.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -1582,14 +1781,18 @@ public async Task<TileMetadataSyncResult> DownloadMetadataForTileAsync(ModelTile
...
         var options = new SyncOptions(
             new HashSet<SyncStepKind>
             {
                 SyncStepKind.IdentifyModel, SyncStepKind.FetchTags, SyncStepKind.FetchImages,
                 SyncStepKind.Thumbnails,
             },
             ForceIdentify: true,
-            ForceThumbnails: true);
+            ForceThumbnails: true,
+            RetryPolicy: _scrollRetryPolicy);
```
Little-God1983: **Two of the three thumbnail paths ignore the user's `SyncThumbnailConcurrency` — but `Doc/LoraViewer.md` in this PR says they honour it.**

This per-tile `SyncOptions` omits `ThumbnailConcurrency` (so it gets the record default, 4), and `CivitaiModelDownloader.RunCompletionSyncAsync` (`CivitaiModelDownloader.cs:434`) omits both `ThumbnailConcurrency` **and** `RetryPolicy`. Yet the doc added in this PR states:

> The same clamp applies wherever a model has more than one due image at once: the downloader's post-download completion sync and a per-tile sync both fetch that model's outstanding thumbnails through the same bounded path…

and, for the windows, that the viewer "hands it to the bulk run, the per-LoRA button and the tiles alike".

Failure scenario: a user on a metered/throttled connection sets **Thumbnail downloads in parallel = 1**. Every model download's completion sync and every per-tile "Download metadata" still fires up to 4 concurrent CDN GETs, and the completion sync additionally judges due-ness by `SyncRetryPolicy.Default` rather than the user's error window. The setting silently applies to one of three paths.

Suggested fix: pass the settings-derived `ThumbnailConcurrency` here, and give `RunCompletionSyncAsync` the same settings-derived `SyncOptions` (or fix the doc to say only the bulk run is configurable).

DiffusionNexus.Domain/Models/SettingsExportData.cs
```diff
@@ -50,6 +50,21 @@ public sealed record SettingsExportData
     public bool MergeLoraSources { get; init; }
     public int LoraUpdateCheckStalenessDays { get; init; } = 3;
 
+    // ?? Metadata Sync ????????????????????????????????????????
+
+    /// <summary>Days before the bulk metadata sync re-checks a not-identified model. Default 30.</summary>
+    public int SyncNotIdentifiedRetryDays { get; init; } = 30;
```
Little-God1983: **Three fields added to `SettingsExportData` without bumping `SettingsExportSchema.CurrentVersion` — against the rule stated in this very file.**

`SettingsExportSchema` (line 152-163 of this file) documents: *"Bump `CurrentVersion` whenever new fields are added to `SettingsExportData` so the importer can detect older formats."* `CurrentVersion` is still `2`.

Failure scenario: a v2 file written before this PR and a v2 file written after it are indistinguishable, so `ImportAsync` cannot tell "the field is absent because the file predates the feature" from "the user explicitly chose 30/1/4". Any future migration or diagnostic that needs that distinction (as `BackupDatasetImagesEnabled ?? LegacyAutoBackupEnabled` needed it for the v1→v2 split) has no signal to work from. The PR body notes this as "unavoidable for newly added fields" — a version bump is exactly the mechanism the file provides to make it avoidable.

Suggested fix: `CurrentVersion = 3` with a one-line note ("v3: metadata-sync retry windows + thumbnail concurrency").

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -930,6 +1101,34 @@ private async Task DownloadMissingMetadataAsync()
...
+    /// the counts it shows include files added since the app started.
+    /// </summary>
+    private static readonly IReadOnlySet<SyncStepKind> DiscoverOnly =
+        new HashSet<SyncStepKind> { SyncStepKind.DiscoverFiles };
+
+    /// <summary>
+    /// The four steps the plan dialog offers. Discovery is not among them: by the time the dialog
+    /// opens it has already happened, and its count could never have been shown in advance anyway.
+    /// </summary>
+    private static readonly IReadOnlySet<SyncStepKind> PlannedStepKinds = new HashSet<SyncStepKind>
```
Little-God1983: **The set of "everything except discovery" is now hardcoded in two places, replacing a definition derived from the enum.**

`PlannedStepKinds` here and `SyncPlanDialogViewModel.RowOrder` (+ `AllRowKinds`) each spell out the same four `SyncStepKind` values. The code they replace used `SyncOptions.All`, which was `new HashSet<SyncStepKind>(Enum.GetValues<SyncStepKind>())` — self-maintaining.

Failure cost: register a fifth step (the design doc already anticipates more) and it is silently absent from the bulk button's plan *and* from the dialog, with no compile error and no test failure — the plan simply never contains it and nobody notices until a user reports missing data. Two lists must now be kept in step with the enum and with each other.

Suggested fix: derive both from the enum, e.g. a single `SyncOptions.AllExcept(SyncStepKind.DiscoverFiles)` (or `SyncStepKind` values minus discovery) and have `RowOrder` order that set rather than redefine it.

DiffusionNexus.UI/ViewModels/SyncPlanDialogViewModel.cs
```diff
@@ -0,0 +1,268 @@
...
+    internal Action? SelectionChanged { get; set; }
+}
+
+/// <summary>
+/// Confirmation dialog for a metadata sync run: the four candidate steps with their counts,
+/// the force toggles that re-plan on the spot, and the options the run is started with.
+/// </summary>
+public sealed partial class SyncPlanDialogViewModel : ObservableObject
+{
+    internal const string UpToDateMessage = "Library is up to date — nothing to do";
```
Little-God1983: **Duplicate of `LoraViewerViewModel.UpToDateStatus` (`LoraViewerViewModel.cs:1337`) — the same user-visible string is now defined in two files.**

Both are exactly `"Library is up to date — nothing to do"`, and both are asserted verbatim in tests (`SyncPlanDialogViewModelTests.AllZeroPlan_IsUpToDate_WithNoStart_AndShowsTheLastRun` and `LoraViewerViewModelSyncTests`). The two consts are also *reachable from the same screen in the same second*: the dialog shows one and the status bar behind it shows the other.

Failure cost: reword one (e.g. drop the em dash) and the app shows two different phrasings of the same verdict, with green tests on both sides.

Suggested fix: keep one const and reference it from the other (or move it to `SyncReport`/a shared strings type, next to `SyncReport.Label`, which both view models already share).

DiffusionNexus.UI/ViewModels/SyncReportDialogViewModel.cs
```diff
@@ -0,0 +1,112 @@
...
+        HasFailures = FailureGroups.Count > 0;
+
+        IsPartial = report.Cancelled;
+        PartialText = "Cancelled — partial run. Completed items are recorded and will not be redone.";
+
+        SummaryText = report.Summary;
+        ElapsedText = SyncPlanDialogViewModel.FormatDuration(report.Elapsed);
+
+        HasDiscoveredFiles = newFilesDiscovered > 0;
+        DiscoveredText = newFilesDiscovered switch
```
Little-God1983: **The discovered-count text is a verbatim copy of `SyncPlanDialogViewModel`'s, and this class reaches into that view model for its duration formatter.**

Lines 73-79 here are character-for-character identical to `SyncPlanDialogViewModel` lines 94-101 (`HasDiscoveredFiles` + the `<= 0 / 1 / _` switch), and line 71 calls `SyncPlanDialogViewModel.FormatDuration` — a report view model depending on the *plan* view model purely to borrow a static string helper.

Failure cost: the two dialogs are shown back-to-back in the same flow with the same number, so a change to one phrasing silently desynchronises them; and the cross-view-model static reference makes `SyncPlanDialogViewModel` un-deletable/un-movable without touching the report.

Suggested fix: one small shared helper (e.g. `SyncTextFormatting` next to `SyncReport.Label`, which both already use) holding `DescribeDiscovered(int)` and the duration formatters, and have both view models call it.

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -857,32 +908,144 @@ private async Task DownloadMissingMetadataAsync()
...
+
+            var chosen = choice.Options;
+            _logger?.Info(LogCategory.Network, "CivitaiSync",
+                $"User started sync: steps [{string.Join(", ", chosen.Steps)}] forces " +
+                $"[I:{chosen.ForceIdentify} T:{chosen.ForceTags} Im:{chosen.ForceImages} Th:{chosen.ForceThumbnails}]");
+
+            // Re-planned, not the dialog's plan: it is cheap, the dialog may have been open for
+            // minutes, and the ticks and forces the user came back with select a different set of
+            // items than the one the dialog was built from.
+            var runPlan = await Task.Run(() => _librarySync.PlanAsync(SyncScope.Library, chosen, ct), ct);
```
Little-God1983: **A full four-step selection pass is repeated here even when the dialog just planned the identical options.**

`PlanAsync` runs `SelectAsync` on every requested step, and each of those is a repository query over the whole library plus per-item retry-policy filtering. This button now pays for that pass at least twice (line 950 for the dialog, again here) — three times or more if the user toggled any Force, because each toggle already re-planned with the exact options `BuildResult()` then returns.

Failure cost: on a large library that is a duplicated multi-second DB pass on the UI-blocking critical path of the button, entirely wasted whenever the user changed nothing between the dialog opening and pressing Start (the common case — every row is pre-ticked).

Suggested fix: have `SyncPlanDialogViewModel` hand back the plan it last computed alongside the options, and re-plan here only when the returned options differ from the ones that plan was built with. (`RunStepAsync` already re-selects per step at execution time, so the plan's freshness is only used for the counts.)

DiffusionNexus.UI/ViewModels/LoraViewerViewModel.cs
```diff
@@ -494,6 +508,10 @@ public LoraViewerViewModel(
 
         _ = InitializeBaseModelFilterAsync();
         _ = LoadDestinationFoldersAsync(destination);
+
+        // Once, at startup: the tiles' scroll-fetch gate needs the user's retry windows before the
+        // first bulk sync of the session ever happens, and it must not read settings per tile.
+        ScrollRetryPolicyLoad = LoadScrollRetryPolicyAsync();
```
Little-God1983: **A third startup DB round-trip, added to the constructor without `Task.Run` — and `GetSettingsAsync` is far from a cheap read.**

`LoadScrollRetryPolicyAsync` needs exactly two `int`s, but `AppSettingsService.GetSettingsAsync` does `ClearChangeTracker()` → `GetSettingsWithIncludesAsync` (LoRA sources, image galleries, base-model folders, dataset categories) → `SaveChangesAsync` → `RemoveDuplicateCategoriesAsync` → `RemoveDuplicateFolderRowsAsync`. That is a full graph load plus up to three writes.

Failure cost: SQLite has no true async (the reason every other DB touch in this file is wrapped in `Task.Run`, per the R7 comments at lines 916 and 1050), so this whole thing runs **inline on the UI thread** before the constructor's first real yield — a startup hitch on exactly the view the UI-performance work targeted, and a DB *write* during viewer construction.

Suggested fix: `Task.Run(LoadScrollRetryPolicyAsync)` at minimum, and ideally a narrow read for the two columns rather than the full settings graph.

DiffusionNexus.Domain/Services/Sync/SyncRetryPolicy.cs
```diff
@@ -8,6 +8,17 @@ public sealed record SyncRetryPolicy(TimeSpan NotIdentifiedRetryAfter, TimeSpan
 {
     public static SyncRetryPolicy Default { get; } = new(TimeSpan.FromDays(30), TimeSpan.FromDays(1), 3);
 
+    /// <summary>
+    /// Builds a policy from the user's saved settings. Floors both windows at one day: zero would
+    /// mean "always due", which turns every plan into a full re-run — when that is wanted it is a
+    /// Force checkbox on the plan dialog, not a saved setting. <see cref="MaxErrorAttempts"/> stays
+    /// at the default: it caps API hammering inside one run window and is not a user-facing knob.
+    /// </summary>
+    public static SyncRetryPolicy FromDays(int notIdentifiedDays, int errorDays) =>
```
Little-God1983: **`FromDays` floors both windows but never caps them, so a large stored value throws instead of degrading.**

`TimeSpan.FromDays` throws `OverflowException` above ~10 675 199 days. `Math.Max(1, …)` guards the bottom end only, and the values reaching here are not guaranteed to come from the settings combo boxes: `SettingsExportService.ImportAsync` copies `export.SyncNotIdentifiedRetryDays` / `SyncErrorRetryDays` straight from a JSON file with no range check, and `AppSettingsService.SaveSettingsAsync` persists them unvalidated.

Failure scenario: a hand-edited or corrupted `settings.json` carrying `"syncNotIdentifiedRetryDays": 2147483647` is imported → every subsequent **Download Metadata** press dies at this line and reports "Sync error: TimeSpan overflowed because the duration is too long.", and `LoadScrollRetryPolicyAsync` silently falls back to the default — the button is unusable with no way for the user to see why.

Suggested fix: `Math.Clamp(days, 1, 3650)` on both arguments — the same "the service owns the sane range" stance already applied to `ThumbnailConcurrency` in `LibrarySyncService`.

Little-God1983:
## Review round 1 — all 15 findings verified real, all 15 fixed

Fixed in `aae1dc13..a18657f6` (9 commits, pushed). Suite **4,770 passed / 0 failed / 2 skipped** (+45 tests over the pre-review branch), UI Release build 0 warnings. The wave got an independent adversarial re-review before the push: every finding confirmed structurally fixed against the post-change code, both my refinements RED-verified in both directions, no regressions found.

By finding id:

- **3855607077** (cancel leaves discovered files invisible): `discovered`/`rebuilt` hoisted above the try; a guarded backstop rebuild in the `finally` covers every exit — dialog cancel, missing dialog service, both already-running refusals, and both catch paths. The cancel status now says "Sync cancelled — the scan added N new files." instead of the false "nothing was run".
- **3855607244** (scan failures thrown away): the rebuilt run report now folds in the scan's `Failures`, `UnexpectedFailures`, and `FirstUnexpectedError`; scan failures appear in the report dialog as their own group and in the "· N failed" status count. A cancel-at-dialog with scan failures says so in the status line.
- **3855607429** (IOE catch too wide): the gate now throws a dedicated `SyncAlreadyRunningException` (derives from IOE, so nothing downstream breaks); both catch sites narrowed. A DI/LINQ `InvalidOperationException` now surfaces as "Sync error: …" with an Error log — a new test pins exactly that.
- **3855608567** (Start pressable between queued re-plans): the pending-re-plan counter is raised synchronously at toggle time and only the last link out lowers `IsReplanning`, so `CanStart` stays false across the whole chain. Pinned with a TCS-gated two-toggle test.
- **3855608757** (elapsed excludes the scan + `~` estimate formatter): the scan's `Elapsed` is summed into the run report, and the report renders it through a new exact `FormatElapsed` ("3 min 42 s", no tilde).
- **3855608888** (subset run stamps "Last full sync"): the stamp is now gated — with a refinement on your suggested fix: `chosen.Steps.SetEquals(all four)` would never stamp again once any step legitimately has zero items (BuildResult drops zero-count kinds), so the guard judges the *dialog's rows*: every kind **with work** must have run; a kind with nothing to do is covered by doing nothing. RED-verified in both directions (unticked-with-work → no stamp; zero-count kind absent → still stamps).
- **3855609020** (cancel status from the pre-force plan): wording now reads `dialogVm.IsUpToDate` — the live post-re-plan state.
- **3855609167** (two of three thumbnail paths ignore the settings): all three now honor them — per-tile passes the cached settings concurrency, and `CivitaiModelDownloader`'s completion sync reads settings once per completion (fail-open to defaults, off the transfer path) for both `RetryPolicy` and `ThumbnailConcurrency`. This also closes item 1 of #532. The doc claims you quoted are now true as written.
- **3855610267** (schema version not bumped): `CurrentVersion = 3` with the v3 note. You were right and our earlier "additive convention" reading was wrong — the file's own rule wins; the previous additive field (`09d0ec38`) simply violated it. The importer only enforces a lower bound, so v1/v2 files still import and older builds ignore the new fields.
- **3855610411** (hardcoded four-kind set ×2): `SyncOptions.AllStepsExcept(DiscoverFiles)` derives the viewer's set; the dialog's `RowOrder` derives from the enum (declaration order is display order) — a fifth step now flows into both surfaces automatically, with a test pinning the derivation. Deliberately *not* applied to the per-tile set: auto-opting a single-model fetch into future steps is a decision, not a default.
- **3855610563 / 3855610726** (duplicated copy + cross-VM formatter reach): new `SyncCopy` holds the up-to-date string (exactly one occurrence in production now), `DescribeDiscovered`, and both formatters; the report VM no longer references the plan VM.
- **3855610878** (duplicated selection pass on Start): the dialog hands back the plan it was showing (filtered to ticked kinds, re-labelled with the chosen options) whenever it can vouch for its counts; a failed re-plan returns `Plan = null` and the viewer plans fresh. `RunStepAsync` re-selects per step at execute anyway, so the reused plan is behaviorally identical — the counts only feed the report's Planned column.
- **3855611013** (inline settings read in the ctor): `Task.Run`, aligned with the file's R7 pattern.
- **3855611154** (`FromDays` overflow): `Math.Clamp(days, 1, 3650)` on both windows, theory extended with `int.MaxValue`.

One additional hardening from the re-review: the sync-running flag now drops only **after** the backstop rebuild, so the per-tile button can't briefly re-enable against a tile set mid-swap (`a18657f6`).

Known residuals, named rather than hidden: with plan reuse, a run whose work was done by something else in the dialog-open window reports "Identified 0/3" instead of the up-to-date wording (the plan's counts are cosmetic; the honest number wins); on cancel, "the scan added N files" takes priority over a simultaneous scan-failure message (deliberate ordering); the per-tile step set stays literal (see above).

Follow-up issues from this round's context: #531 (the pre-existing LastBackupAt-nulled-on-save bug), #532 (item 1 closed by this wave; item 2 — the per-tile refusal message — still open).

Little-God1983:
**Post-smoke addition (`5a0b0602`):** the user's first real-library run surfaced a latent Civitai DTO bug — live gallery images carry seeds beyond `Int64.MaxValue` (JS float64 artifacts; a real by-hash payload served `"seed":12859270413054550000`), so `$.images[N].meta.seed` threw and one bad image failed identification of the whole model version. Fix: `TolerantNullableInt64JsonConverter` on `CivitaiImageMeta.Seed` (unusable shapes → null, never clamped), RED-proven with the verbatim payload, 9 new tests. The second DTO family has no image `meta` object, so this is the single fix point.

Also fixed from the same smoke: the report dialog's failure reasons clipped at the window edge (horizontal StackPanel measures with unbounded width, so `TextWrapping` never engaged) — now a bounded grid column that wraps inside the scroll viewer.

Suite: 4,779 passed / 0 failed / 2 skipped.

PR #543, opened 2026-08-27
fix: preserve LastBackupAt across Settings saves (#531) + friendly per-tile sync refusal (#532)

### LINKED ISSUE #531: Settings-page save silently nulls LastBackupAt in the database
## What

Every save from the Settings page nulls `AppSettings.LastBackupAt` in the DB.

Found (and verified real) during the Plan E review on PR #530; **predates that branch** — it exists on `develop` today.

## Mechanism (two layers)

1. `SettingsViewModel` never reads or carries `LastBackupAt`: it is absent from `LoadAsync` and from the detached `new AppSettings { ... }` snapshot the save command builds — so the snapshot carries the CLR default `null`.
2. `AppSettingsService.SaveSettingsAsync` (`AppSettingsService.cs:305`) blind-copies that snapshot value onto the tracked entity: `existingSettings.LastBackupAt = settings.LastBackupAt;`.

Net effect: open Settings, press Save → the recorded last-backup timestamp is gone. Anything that reads it to decide backup staleness sees "never backed up".

## Fix options

- Preferred: remove `LastBackupAt` from the `SaveSettingsAsync` scalar-copy whitelist entirely — it has a dedicated writer (`UpdateLastBackupAtAsync`) and is machine-local bookkeeping, exactly like `LastLibrarySyncAt`, which Plan E deliberately kept out of the whitelist and protected with a capture/replay in the VM plus an e2e preservation test (`AppSettingsServiceSyncSettingsTests`). Mirror that pattern.
- Alternative: capture/replay in the VM like `_loadedLastLibrarySyncAt` — weaker, because every other `SaveSettingsAsync` caller building a detached snapshot re-introduces the bug.

Add a regression test mirroring `ExportImport_DoesNotDisturbLastLibrarySyncAt` / `SavingSettings_PreservesLastLibrarySyncAt`.

## References

- PR #530 "Pre-existing bug found" section
- The protected sibling implementation: `LastLibrarySyncAt` handling in `SettingsViewModel` + `SettingsExportService` (Plan E, branch `feature/metadata-sync-overhaul-e-sync-ui`)

### LINKED ISSUE #532: Sync consistency follow-ups from the PR #530 final review
Two small, non-blocking consistency items noted by the Plan E final whole-branch review (PR #530). Neither is a live defect today; both are worth closing so the sync surface stays coherent.

## 1. The downloader's post-download completion sync ignores the user's retry windows

`CivitaiModelDownloader.RunCompletionSyncAsync` (`CivitaiModelDownloader.cs:~434`) builds its `SyncOptions` without a `RetryPolicy`, so it judges due-ness by `SyncRetryPolicy.Default` — it is now the **only** sync entry point not using the settings-derived windows (the bulk flow, the per-tile button, and the scroll gate all do since Plan E).

Currently inert: a just-downloaded model's `CheckedAt` stamps are null, so every step is due under any window. But the inconsistency is a trap for the next change that makes the completion sync re-visit older rows.

Fix: resolve the policy from settings (`SyncRetryPolicy.FromDays(settings.SyncNotIdentifiedRetryDays, settings.SyncErrorRetryDays)`) the same way `LoraViewerViewModel` does, or accept an injected policy provider.

## 2. Per-tile "Download Metadata" can surface the single-flight refusal as an error

`LoraViewerViewModel.DownloadMetadataForTileAsync` (~line 1806) does not wrap `ExecuteAsync` for the gate's `InvalidOperationException`. If a post-download completion sync grabs the single-flight gate between the entry guard and the execute, the detail panel shows "Metadata download failed: A library sync is already running." with an Error log, instead of the friendly refusal the bulk path gives (`AlreadyRunningStatus`, Info log). Pre-existing WP2 behavior; the caller catches it safely, so this is message quality, not a crash.

Fix: mirror the bulk path's IOE handling (catch → status "A metadata sync is already running.", Info log, no error row).

Context: both items are documented in the "Notes from the final whole-branch review" section of PR #530.

Little-God1983:
Closes #531. Closes #532.

The two spin-offs left over from the metadata-sync overhaul (#521), one commit each, TDD throughout — every fix was watched failing first. Suite **4815 / 0 / 2** (+6 tests), IntegrationTests 2/2, UI build 0 warnings, all in Debug (the user's running app locks `bin\Release`; CI builds Release).

## #531 — every Settings save nulled `LastBackupAt` (`c6214464`)

`LastBackupAt` is machine-local bookkeeping with a dedicated writer (`UpdateLastBackupAtAsync`), but it sat in `SaveSettingsAsync`'s scalar whitelist. Nothing that builds one of those detached `AppSettings` snapshots has a value for it — the Settings page has no such field, and the export DTO doesn't carry it — so every save copied the CLR default `null` over the stored timestamp. `BackupScheduler` then read "never backed up" and treated a fresh backup as overdue.

Fixed by removing the column from the whitelist (the issue's preferred option) rather than carrying it through the Settings ViewModel: the guarantee then belongs to the service and holds however the caller built its snapshot — which also closes the same hole in `SettingsExportService.ImportAsync`, a second door the capture/replay alternative would have left open. Same rule and reasoning as `LastLibrarySyncAt`, which Plan E kept out of the whitelist for exactly this reason.

Checked before removing it: the only other caller that saves a snapshot carrying this column (`OutputsFolderRegistrar`) saves a *loaded* entity, so it is unaffected, and `LastBackupAt`/`LastLibrarySyncAt` are the only two dedicated-stamp columns in `IAppSettingsService` — the whitelist has no others of this kind left.

`AppSettingsServiceBackupStampTests` (real SQLite, mirroring the `LastLibrarySyncAt` pair): a Settings-page-shaped save preserves the stamp (RED-verified), an export/import round trip preserves it (RED-verified), and the dedicated writer still moves it forward (guards against over-fixing into a frozen column).

## #532 — the per-tile refusal surfaced as an error (`e74b377c`)

**Item 1 was already closed** by the Plan E fix wave — `BuildCompletionOptionsAsync` resolves the retry policy from settings. Verified in code before writing anything; nothing to do.

**Item 2:** `DownloadMetadataForTileAsync` guarded on its own `SyncInFlight` flag but never caught the gate's `SyncAlreadyRunningException`, and the gate is taken *inside* the service — a download's completion sync can grab it in the window between the two. The refusal escaped into the detail panel's generic catch as "Metadata download failed: A library sync is already running." with an Error log and a stack trace, for a run that simply has to wait its turn.

Now caught and reported exactly as the bulk path reports it (`ReportServiceAlreadyRunning`: status line + Info log, no error row). One addition beyond the issue text: the outcome carries `Refused`, because without it the refusal fell through to **"Nothing to refresh for this model."** — a claim about the model, made when the truth is that the button was busy. The pre-existing entry-guard door returns the same `Refused` outcome for the same reason, so both doors give one answer.

Three tests, all RED-verified (the first failed with the escaping `SyncAlreadyRunningException` at `LoraViewerViewModel.cs:2068`, exactly as the issue predicted): the gate refusal is not a failure and states itself, the entry-guard refusal is the same answer, and a refused run leaves the in-flight flag down so the next press is not refused by our own guard.

`Doc/LoraViewer.md` updated in both places the contract is taught — the per-tile message order and the fallbacks table.

## Not done here

`ModelFileSyncService.FullSyncAsync` still discards `DiscoveryResult.RepointedCount` (noted in the #542 review). It has no callers anywhere in the solution — dead code, and deleting it is a separate decision.

### DIFF
--- a/DiffusionNexus.Service/Services/AppSettingsService.cs
+++ b/DiffusionNexus.Service/Services/AppSettingsService.cs
@@ -302,7 +302,13 @@ public sealed class AppSettingsService : IAppSettingsService
         existingSettings.AutoBackupIntervalHours = settings.AutoBackupIntervalHours;
         existingSettings.AutoBackupLocation = settings.AutoBackupLocation;
         existingSettings.MaxBackups = settings.MaxBackups;
-        existingSettings.LastBackupAt = settings.LastBackupAt;
+        // LastBackupAt is deliberately NOT copied (#531). It is machine-local bookkeeping written
+        // only by UpdateLastBackupAtAsync, and nothing that builds one of these detached snapshots
+        // — the Settings page, SettingsExportService.ImportAsync — has a value for it to carry, so
+        // copying it here wrote the CLR default null over the real stamp on every save: the backup
+        // scheduler then read "never backed up". The freshly loaded tracked entity above already
+        // holds the true value; leaving it alone is the fix. Same rule as LastLibrarySyncAt, which
+        // is absent from this whitelist for exactly the same reason.
         existingSettings.ComfyUiServerUrl = settings.ComfyUiServerUrl;
         existingSettings.LoraUpdateCheckStalenessDays = settings.LoraUpdateCheckStalenessDays;
         existingSettings.SyncNotIdentifiedRetryDays = settings.SyncNotIdentifiedRetryDays;
--- a/DiffusionNexus.Tests/Service/Services/AppSettingsServiceBackupStampTests.cs
+++ b/DiffusionNexus.Tests/Service/Services/AppSettingsServiceBackupStampTests.cs
@@ -1,0 +1,164 @@
+using DiffusionNexus.DataAccess;
+using DiffusionNexus.Domain.Entities;
+using DiffusionNexus.Domain.Services;
+using DiffusionNexus.Service.Services;
+using FluentAssertions;
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Moq;
+
+namespace DiffusionNexus.Tests.Service.Services;
+
+/// <summary>
+/// #531. <see cref="AppSettings.LastBackupAt"/> is machine-local bookkeeping with a dedicated
+/// writer (<see cref="IAppSettingsService.UpdateLastBackupAtAsync"/>) — nothing on the Settings
+/// page edits it, so no detached snapshot a caller builds can carry a real value for it. While it
+/// sat in <c>SaveSettingsAsync</c>'s scalar whitelist, every such save copied the snapshot's CLR
+/// default <c>null</c> over the stored timestamp: open Settings, press Save, and the record of the
+/// last backup was gone — <see cref="DiffusionNexus.Service.Services.BackupScheduler"/> then read
+/// "never backed up" and treated a fresh backup as overdue.
+/// <para>
+/// The guarantee is the service's, not any one caller's: it holds however the snapshot was built,
+/// which is why the column was removed from the whitelist rather than carried through the Settings
+/// ViewModel. Same reasoning, same shape as the <see cref="AppSettings.LastLibrarySyncAt"/> pair in
+/// <c>AppSettingsServiceSyncSettingsTests</c>.
+/// </para>
+/// </summary>
+public sealed class AppSettingsServiceBackupStampTests : IDisposable
+{
+    private readonly SqliteConnection _connection;
+    private readonly ServiceProvider _serviceProvider;
+    private readonly DirectoryInfo _tempDir = Directory.CreateTempSubdirectory("dn-appsettings-backup-tests-");
+
+    public AppSettingsServiceBackupStampTests()
+    {
+        _connection = new SqliteConnection("DataSource=:memory:");
+        _connection.Open();
+
+        var secureStorageMock = new Mock<ISecureStorage>();
+        secureStorageMock.Setup(s => s.Encrypt(It.IsAny<string?>())).Returns<string?>(v => v);
+        secureStorageMock.Setup(s => s.Decrypt(It.IsAny<string?>())).Returns<string?>(v => v);
+
+        var services = new ServiceCollection();
+        services.AddDataAccessLayer(options => options.UseSqlite(_connection));
+        services.AddSingleton(secureStorageMock.Object);
+        services.AddTransient<IAppSettingsService, AppSettingsService>();
+        services.AddTransient<ISettingsExportService, SettingsExportService>();
+
+        _serviceProvider = services.BuildServiceProvider();
+
+        using var scope = _serviceProvider.CreateScope();
+        var context = scope.ServiceProvider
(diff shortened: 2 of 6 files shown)


Little-God1983:
## Review — no blocking findings

Both non-test hunks verified against the surrounding code, not just the diff.

### #531 — `LastBackupAt` whitelist removal

Sound, and the reasoning in the description checks out:

- Grepped every write of the column. The only producers are `UpdateLastBackupAtAsync` (called by `BackupScheduler` and `DatasetBackupService`) and the removed whitelist line. The only consumer is `BackupScheduler.cs:98`. Nothing builds a detached `AppSettings` carrying a real value for it, so removing the copy cannot lose a write.
- Both snapshot builders confirmed: `SettingsViewModel.cs:759` and `SettingsExportService` `ImportAsync` — neither sets it. `OutputsFolderRegistrar.cs:61` saves a loaded entity, so it's unaffected either way.
- I did chase the obvious adjacent suspect — `existingSettings.UpdatedAt = settings.UpdatedAt;` on the very next line, same class of "detached snapshot carries a CLR default" hazard, and none of the three callers set `UpdatedAt`. It's a false alarm: `SaveSettingsAsync` stamps `settings.UpdatedAt = DateTimeOffset.UtcNow` at line 222 before the copy block. Noting it so the next reader doesn't have to re-derive it.

### #532 — `Refused` outcome

Also correct:

- The new `catch (SyncAlreadyRunningException)` precedes no other catch on that `try` (it was `try`/`finally` before), so it doesn't shadow anything, and the `finally` still clears `_localSyncActive` + calls `RefreshSyncRunning()` on the refusal path.
- `Faulted` is guarded on `Report is not null`, and `AlreadyRunning` carries `Report: null`, so the new `else if (outcome.Refused)` branch in `OnDetailMetadataDownloadRequested` is genuinely reachable — ordered after `Faulted` and before the `IdentifyPlanned` fallback, which is the right precedence.
- `TileMetadataSyncResult` has no consumers outside the VM and its tests, so widening the record with a defaulted `Refused` breaks nothing.
- The entry-guard return changing `NotRun` → `AlreadyRunning` is safe for the same reason (no caller compares against `NotRun` by value).

### Two nits, neither blocking

1. **Stale-ish comment in `SettingsExportService.ImportAsync`** (~line 168): it explains that `LastLibrarySyncAt` is deliberately left unset because "`SaveSettingsAsync`'s whitelist does not touch that column". That statement now covers `LastBackupAt` too, and this is exactly the spot where a future editor would helpfully add `LastBackupAt = export...`. Worth naming both columns there.

2. **`DownloadMetadataForTile_ARefusedRunStillClearsTheInFlightFlag` asserts slightly more than it proves.** `IsSyncRunning` is `SyncInFlight` = `_localSyncActive || _metadataSyncCts is not null || _librarySync?.IsRunning`. The assertion passes because the mocked `ILibrarySyncService.IsRunning` returns `false` by default. In the real scenario this test describes — the gate refused us because *another* run holds it — `_librarySync.IsRunning` would still be `true` and `IsSyncRunning` would correctly stay `true`. The test does verify the thing that matters (`_localSyncActive` is cleared), but the name and the "the button must come back" reason read as a stronger claim than the mock actually exercises. Setting up `IsRunning => true` and asserting on the flag itself would say what it means.

CI green (`build` pass). Nothing here needs a change before merge.

PR #544, opened 2026-08-27
feat(sorter): preview shows the move, not just the destination

Little-God1983:
The LoRA Sorter's preview pane drew the **destination** layout only. It could tell you a folder was about to hold twelve files, but not which file was leaving where — which is the question you have to answer before letting something move bytes on disk.

It is now a before → after. Design was mocked up and approved before implementation: https://claude.ai/code/artifact/0dc1802f-a146-48b4-9b25-7985f9bfbfd4

One commit, TDD throughout — all 11 new tests watched failing first. Suite **4826 / 0 / 2** (+11), UI build 0 warnings, Debug (the running app locks `bin\Release`; CI builds Release).

## Two trees, one plan

`SOURCE (NOW)` on the left, `AFTER SORTING` on the right, `GridSplitter` between them.

Both are built by one pass over the same `LoraSortPlan`, read from opposite ends — `PlannedMove.Candidate.FilePath` builds the left, `PlannedMove.TargetFilePath` the right. That is the load-bearing part: the source side is not re-derived and never goes back to disk, so no option toggle can leave the two halves describing different plans, and re-planning rebuilds both together.

## Click to link

`SelectPreviewNodeCommand`. Every `PlannedMove` carries an index; a file node holds its own, a folder node the union of everything beneath it — so a folder click lights every destination its files reach.

- **Only file rows light.** Lighting destination folders too would mean a folder click paints most of the other pane, which says "somewhere over there" rather than "these rows". The folders on the path are expanded instead — a highlight inside a collapsed folder is a highlight nobody sees.
- **Exactly one counterpart is the scroll target** (`IsPrimaryLink`, first in tree order). A folder click can light a dozen rows; scrolling to all of them means scrolling to none.
- **A second click clears it**, and so does a re-plan — the nodes a highlight referred to stop existing when the trees are rebuilt.

## What only the source side shows

Skipped duplicates. The destination tree drops them because nothing arrives, but the file is still sitting in the user's folder and this is the only side that can say so — struck through, reading *duplicate — skipped*. Source rows also carry `already here`, `renamed on arrival`, and per folder a count of departures.

That count reads **"all 7 leave"** and deliberately never *"empties"*. The plan covers model files and their sidecars, not whatever else is in that folder, and **Delete empty source folders** removes a directory only when it is genuinely empty at execution time. Pinned by a test that asserts the string does not contain "empt".

Kind chips stay on the destination side — the question they answer is what a folder is about to *receive*. The ✓ / ~ / ✗ identity mark appears on both, because the source side is where "why is this one heading for Unknown?" actually gets asked.

## Two deviations from the approved mockup, both small

1. **One `DataTemplate`, not a subclass per side.** The design said two node subclasses so the view could template each side by type. Once `Note` was a plain string and the chips were a visibility binding, the subclasses earned nothing but a DataTemplate-ordering hazard — Avalonia matches templates by assignability, so a base-typed template silently claims derived nodes. One template with two bindings (`!IsSourceSide` for chips, `Note` for the line) does the same job.
2. **A skipped duplicate keeps its ✓ / ~ / ✗ mark.** The mockup showed those rows unmarked. Suppressing it would have made the folder rollup — which counts the file either way — say something its own rows could not explain.

## Notes for review

- `SortPreviewNodeViewModel.Note` and the folder counters raise no change notification, like the counters that were already there. Both rely on the build running synchronously on the UI thread, so the first layout pass cannot happen until every value is final. That invariant is now written down in a `<remarks>` on `AnnotateSourceFolders`, because this change adds a second after-the-fact mutation depending on it — move any of it past an `await` and rows render their defaults.
- `LightCounterparts` asks `moveIds.Overlaps(node.MoveIds)`, not the mirror form. `Overlaps` enumerates its argument and probes the receiver, so a file node's single id is one lookup; the other way round walks the whole selection once per file in the tree.
- Scroll-into-view is in the view's code-behind rather than a new attached property — the codebase has no attached-behavior precedent, and this is view behavior triggered by exactly one thing. It is posted at `Background` priority because selecting expands folders, and until that layout pass runs the row has no container or a stale offset.
- Both trees stay realized (no virtualization) — unchanged from the single tree, and a deliberate non-goal here.

## Not in scope

No per-file overrides or exclusions, no drawn connector lines, no virtualization work. `PreviewRoots` keeps its name and meaning, so the ~20 existing assertions against it were untouched.

## Second commit — a search box per pane

Added after smoke-testing the pane on a real 1406-file library: finding one file to check where it lands meant scrolling. Each pane now has its own box in its header.

They filter **independently**, by choice. The two sides are asked different questions — *where is this file now?* against *what is landing in Unknown?* — and a renamed file does not even carry the same name on both of them.

- A **file** survives if its name contains the text. A **folder** survives if its own name matches, in which case everything under it stays, or if anything beneath it does.
- A folder opens **only for a match beneath it**. A match inside a collapsed folder is one nobody sees; a folder that matched on its own name is already the answer to what was typed.
- The header carries a live `3 of 1406 files`, absent when nothing is filtered so an unfiltered pane never reads "1406 of 1406". A pane that matched nothing says *No files match* rather than showing an empty tree under a box with text in it.

Mechanically it walks the nodes that already exist — no re-plan, no disk, no Civitai — so a click-to-link highlight survives typing. Three details that are load-bearing rather than decorative:

1. **Each keystroke re-filters the tree as the user left it**, not as the previous keystroke revealed it. Otherwise `k`, `ke`, `kee` ratchets folders open one at a time and clearing leaves a tree nobody expanded.
2. **The text survives a re-plan** and is re-applied to the fresh tree. An option toggle silently un-filtering the pane someone is reading is the one thing a filter must not do.
3. **`IsPrimaryLink` is only claimed by a visible row.** Independent boxes mean a counterpart can be hidden by the other pane's filter, and a scroll target with no container scrolls nowhere.

Eight tests, all watched failing first — three of them passed against the inert stub and were rewritten until they could not, which is the only reason the "a folder match is still a filter" and "clearing restores the expansion" cases assert anything. `SelfAndDescendants` moved onto the node, replacing the walk the view model kept privately.

No Unified Console logging here, against the standing rule: it would fire per keystroke and record nothing anyone could act on. Say the word if you want it anyway.

## Third commit — open in folder, an order picker, and a straight right edge

**Right-click → Open in folder.** Through the same `IProcessLauncher` seam the generation gallery uses: a file is selected in Explorer, a folder is opened. Every node now carries the path it stands for — where it is now on the source side, where it would be on the destination side. The item greys out when there is nowhere to go, which is the normal state of a destination folder before anything has been sorted; a destination *file* still resolves to the folder it will land in when that folder already exists. Opening the nearest existing ancestor instead was rejected — it takes you somewhere you did not click.

**An order picker in the footer**, driving both trees, because they exist to be compared row against row. The part that is not just a new option: it orders **every level**. Only the roots were ever sorted; below them rows came out in whatever order the plan produced, which is what made a deep tree read as arbitrary. Folders before files, then size biggest-first or name A–Z. Re-ordering moves the existing nodes rather than rebuilding them, so a link highlight and a typed search filter both survive it.

**The ragged right edge was not the badges.** Each level indented the whole row by 18px, so every row's *right* edge marched left as the tree got deeper — the badges were fine, the rows were different widths. The indent moves onto the row's own name (`Depth` → `Indent`), so every row spans the full pane and shares one right edge. The mark then gets a fixed 16px slot, and the count and size fixed right-aligned columns of their own, so a folder's "12 LoRAs" can no longer push its size out of line with the file sizes beneath it.

Eight more tests, all watched failing first — and again three passed against the inert stub (the two negative cases, plus a sort order that happened to match insertion order) and were rewritten until they could not. The context menu uses a `Click` handler rather than the `$parent[ItemsControl]` binding the canvas view uses for its menu: this template is recursive, so the nearest `ItemsControl` above a row is the folder holding it, not the view.

## Fourth commit — the picker made visible, and a Default that undoes

The order picker shipped invisible: `SortOrders` was `static`, and Avalonia binds instance members only — silently, with no binding error. Fixed alongside the second defect the comparison with `CivitaiBrowserView` surfaced (`MinHeight="0"` with no `MinWidth` collapsed the empty ComboBox). The picker also gained **Default** as a real third option: the plan's own order, remembered per node (`Order`) rather than recomputed, so going back to it *undoes* a sort instead of merely stopping sorting.

Disclosure: the two Default-order tests were written and implemented in the same breath — recovered by temporarily neutering the Default branch and watching both go red, but that is a weaker guarantee than red-first.

## Fifth commit — the picker where it is looked for, and hiding what is settled

The picker moved from the footer to a right-aligned toolbar **above both panes** after it was looked for twice and not found. And a new checkbox under the structure radios — **Hide files already in the right folder** — drops every `AlreadyInPlace` move from both trees (both, or the two sides stop describing the same set of moves). View-only by construction: rows are dropped at tree-build time, not out of the plan, so `TransferCount` and what Start actually does are untouched. Folder notes count only the rows still shown, and the summary gains `(hidden)` after the already-in-place count — the one place the real denominator survives.

## Sixth commit — excludable folders ("Never sort this folder")

A hand-curated utility folder — the motivating case is a `Lightning` folder of accelerator LoRAs, which Civitai does not know — was being dismantled by the sorter: scattered by base model when the names happened to be guessable, dumped into `Unknown\` when not. Right-clicking a source-side folder now offers **Never sort this folder**:

- Files under it are partitioned out of the candidates **before** the plan is built and before the name rung gets a say — Start cannot touch them, `TransferCount` ignores them, the history manifest never records them.
- They stay visible in both trees as **dimmed stay-put rows** (synthetic move ids keep click-to-link working; the destination side draws them only when their unchanged location falls under the target root) — a row that silently vanishes reads as a file the sorter lost.
- They show none of the ✓/~/✗ marks (those answer *is the destination known*, and an excluded file has no destination), and their identity rolls up as Identified so an excluded `???` row cannot drag an ✗ up the tree. The exclusion root's note reads *excluded — won't be sorted*; parent notes count them as staying ("1 of 2 leave"); the summary gains `· N excluded`; they survive *Hide files already in the right folder* — hiding noise is one thing, hiding a choice the user made is another.
- Folder dimming is a semantic rollup (every file beneath is excluded), not a path comparison, so a destination folder that merely shares a name with an excluded source folder is never mis-dimmed.
- Persisted as a JSON path list (`AppSettings.LoraSorterExcludedFoldersJson`: new EF migration, recovery-map entry, settings-export mirror) and restored before the first preview. Un-exclude via the context menu on the exact excluded path, or the left rail's **EXCLUDED FOLDERS** ✕ list — visible even when the folder itself is scrolled away. An emptied list persists `null`, not `[]`.

Eleven tests, nine watched red first; the search-filter and hide-settled interplay tests pass against the inert surface by construction and stand as regression guards.

## Owed by the user

GUI smoke on a real library: folder click across a deep tree, the splitter, a source folder that only partly empties, whether the teal highlight reads clearly next to the existing green/amber/red marks, the two search boxes (including whether 150px is enough width at your window size), the right-click menu, the order picker in its new spot above the panes (all three orders, and that Default restores the opening order), the hide-settled checkbox, and the new exclusion flow: right-click your `Lightning` folder → *Never sort this folder*, confirm it dims in both trees and drops out of the transfer count, restart the app to confirm the exclusion survives, and un-exclude from the left rail.


Little-God1983:
## Review — `feature/sorter-preview-before-after`

Full pass over the diff (5 commits, 16 files, ~3.9k lines) plus surrounding context. `dotnet build` on `DiffusionNexus.UI` is clean; the 108 `LoraSorterViewModelTests` pass.

**What holds up.** Both trees really are one pass over one `LoraSortPlan` read from opposite ends, so the halves cannot diverge. Exclusion partitions candidates *before* `ApplyNameGuesses`, so a name guess can never move a parked file — that ordering is the load-bearing bit and it is right. Synthetic move ids start at `plan.Moves.Count`, so they cannot collide with real ones. The non-observable `Note` / `LoraCount` / `IsExcluded` mutations are genuinely safe: `BuildPreviewTree` runs to completion on the UI thread before the first layout pass, and the `<remarks>` now says so. `IsWithin`'s drive-root handling is correct, and the new `AppSettings` column is wired through the migration, the model snapshot, `DatabaseRecoveryService`'s repair map, and *both* directions of `SettingsExportService`. Excluded rows correctly survive hide-settled (separate loop), and the filter's reference-keyed expansion snapshot is sound.

Five findings, none blocking.

---

**1. `MarkExcludedFolders` rolls up from surviving rows, so hide-settled can mark an unexcluded folder as excluded** — `DiffusionNexus.UI/ViewModels/LoraSorterViewModel.cs:1690`

`node.IsExcluded = files.Count > 0 && files.All(f => f.IsExcluded)` runs over the tree *after* `BuildPreviewTree` has already dropped every `AlreadyInPlace` row (line 1602). A folder holding settled files plus an excluded subfolder therefore has only excluded files left in the tree, and gets marked excluded itself.

Consequence: it dims, its identity mark is suppressed, and — because `CanExclude => … && !IsExcluded` — **"Never sort this folder" disappears from its menu**, while `IsExclusionRoot` is false so "Sort this folder again" does not appear either. The folder becomes un-excludable purely because a view checkbox is ticked. (The moved-files framing would be worse still; it isn't that — the hidden rows are no-ops — but the menu loss and the dimming are real.) Roll up over the *plan's* files rather than the surviving rows, or exclude hidden rows from the `All` test.

**2. The un-exclude menu item can be a silent no-op** — `LoraSorterViewModel.cs:1449` vs `:1703`

`IsExclusionRootPath` decides whether to *show* "Sort this folder again" using `GetFullPath` + `TrimEndingDirectorySeparator` + `OrdinalIgnoreCase`. `RemoveExclusionAsync` then looks the entry up with raw `string.Equals(f, path, OrdinalIgnoreCase)` and returns silently on a miss. Any stored path that is equivalent but differently spelled — forward slashes from a yaml-registered source, a trailing separator, a relative segment — shows the menu item and does nothing when clicked. The left rail's ✕ is unaffected (it passes the stored string itself). Reuse the same normalized comparison in `RemoveExclusionAsync`.

**3. Schema version not bumped for the new export field** — `DiffusionNexus.Domain/Models/SettingsExportData.cs:174`

`LoraSorterExcludedFoldersJson` joins the export, but `SettingsExportSchema.CurrentVersion` stays at `3`. The type's own remarks say to bump on every added field, spelling out exactly why: it is the only signal a later migration has for "the file predates the feature" vs "the user chose nothing" — the distinction the v1→v2 `LegacyAutoBackupEnabled` split needed. Bumping is documented as cheap in both directions.

**4. The hide-settled checkbox re-plans, though it is documented as view-only** — `LoraSorterViewModel.cs:367`

`OnIgnoreFilesAlreadyInPlaceChanged` fires the whole `RecomputePreviewAsync`, including `BuildPlan` and its lazy collision hashing — the same path whose own comment notes "planning can run for minutes on a large library". Only `BuildPreviewTree` and the summary string read the flag. The candidate cache saves the enumeration, not the plan. Rebuild the tree from `_lastPlan` instead.

**5. `SettingsExportData.cs` and `SettingsExportService.cs` were rewritten CRLF→LF**

`--numstat` says 180/179 and 243/241 lines. With `--ignore-cr-at-eol` the real change is **1 line and 2 lines**. Every other line is a line-ending flip: blame reset on both files and a guaranteed conflict with any concurrent edit. Restore CRLF and re-apply the three lines.

---

Nothing here touches the design decisions the description defends — the single-pass build, the exclusion ordering, and the "all 7 leave" wording all check out.

Little-God1983:
All five findings verified genuine and fixed in 321f17e9:

1. **Roll-up vs hide-settled** — `BuildPreviewTree` now records the folder chains of the settled rows it hides; `MarkExcludedFolders` spares any folder on that list, so it keeps its menu and stays undimmed on both sides. Red-first test: `HidingSettledFilesDoesNotMarkTheirFolderExcluded`.
2. **Silent un-exclude no-op** — extracted `PathsEqual` (GetFullPath + TrimEndingDirectorySeparator, OrdinalIgnoreCase, literal-compare fallback for unnormalizable entries) and made both `IsExclusionRootPath` and `RemoveExclusionAsync` use it. Test: `RemovingAnExclusionForgivesTheStoredPathSpelling`.
3. **Schema version** — bumped to 4 with the v4 line in the version history; the pinning test updated first.
4. **Hide-settled re-plan** — the toggle now redraws from `_lastPlan`/`_lastTargetRoot`/`_lastExcluded` (cleared wherever the plan is) and re-words the summary via a shared `UpdatePreviewSummary`; same plan reference, so an armed Start stays armed. Test counts planner disk probes: `TogglingHideSettledRebuildsTheTreeWithoutReplanning`.
5. **CRLF** — both files restored; `git diff develop` on them is now 1 + 2 real lines.

Suite 4863 passed / 0 failed; UI Debug build 0 warnings.

PR #546, opened 2026-08-28
fix(civitai-queue): stop a second Start from double-running a gate-waiting job

Little-God1983:
Follow-up review of #545 (already merged).

## The regression

#545 made `StartAllAsync` **join** the current run epoch (`_runCts ??= new`) instead of cancelling it. That correctly fixed Start killing in-flight downloads — but the epoch cancel was also the only thing preventing a job being scheduled twice.

A job parked inside `RunGatedAsync`'s `_gate.WaitAsync` is still `Queued`: `Status = Downloading` is set inside `RunJobAsync`, *after* the slot is taken. So a second `StartAllAsync` re-picks it in `Jobs.Where(j => j.Status == JobStatus.Queued)` and schedules a **second concurrent runner for the same job object**.

- Before #545: the epoch cancel tore the old waiter down, so exactly one runner survived.
- After #545: both survive → two transfers, two `DownloadCoordinator` entries, doubled bandwidth, and a collision-renamed duplicate file on disk.

The worker pool is two wide (`_gate = new(2)`), so **any queue of three or more jobs hits this on a second Start** — including the re-run path #545 itself added.

## Fix

`RunGatedAsync` claims the job in a `_scheduled` set before its first `await` (so the caller's synchronous `Select(...).ToList()` over pending jobs already sees the claim) and releases it in a `finally`. Reference identity; locked, since the release runs on a pool thread. This covers `RetryJobAsync` as well.

## Tests

New `StartAllAsync_DoesNotStartAJobThatIsAlreadyWaitingForAWorkerSlot` — verified to fail on the merged code (`CallsByVersion[3]` was 2, expected 1).

Full suite: **4867 passed, 0 failed, 2 skipped**. `DiffusionNexus.sln` builds clean.

## Rest of #545 reviewed and clean

- `_runCts ??=` epoch join and the `ResetForRetry` re-queue of Cancelled jobs are correct.
- `ClearAll` nulling the fired CTS closes the "new jobs instantly abort" hole; `AbortAllActive` already did the same, so both cancel sites are covered.
- Minor, not changed: `_runCts` is dropped without `Dispose()`. Disposing it while runners still hold linked tokens risks `ObjectDisposedException`, and the linked CTSs are already `using`-disposed, so the leak is a bare CTS with no registrations.

### DIFF
--- a/DiffusionNexus.Tests/Viewer/CivitaiDownloadQueueStartResumeTests.cs
+++ b/DiffusionNexus.Tests/Viewer/CivitaiDownloadQueueStartResumeTests.cs
@@ -79,7 +79,7 @@ public sealed class CivitaiDownloadQueueStartResumeTests : IDisposable
         }
     }
 
-    private CivitaiDownloadQueue Queue(ICivitaiModelDownloader downloader) => new(
+    private CivitaiDownloadQueue Queue(ICivitaiModelDownloader? downloader) => new(
         downloader, logger: null, civitaiClient: null, destination: null,
         persistPathOverride: Path.Combine(_tempDir, $"q-{Guid.NewGuid():N}.json"));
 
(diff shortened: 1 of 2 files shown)


Little-God1983:
Review follow-ups applied in `09216047`.

**The main one wasn't just "log the silent skip" — the strand had a source worth removing.**

The window you identified was real: `StartAllAsync` reset Cancelled jobs to Queued in a pass *before* scheduling, so a job whose previous runner had posted its terminal status but not yet released its claim got reset and then skipped — Queued, no runner. Rather than only logging it, the reset now happens **inside** the claim:

- `pending` selects `Queued or Cancelled` in one pass (this also kills the per-job perf nit — the separate reset loop is gone).
- `RunGatedAsync` does `if (job.Status == JobStatus.Cancelled) job.ResetForRetry();` after `TryClaim` succeeds.

A job that loses the claim now keeps its accurate `Cancelled` status instead of being stranded at `Queued`. Same reasoning in `RetryJobAsync`: claim first, *then* `ResetForRetry`, inside a `try`/`finally`.

**`RetryJobAsync`'s Queued guard is worth more than a defensive nit.** Without it, a Retry on a gate-waiting job awaited `_gate.WaitAsync` for the very slot its own live runner was waiting on — it blocks until a worker frees, then double-runs the job. The new test hangs outright when the guard is removed, which is how I confirmed it.

**Also applied:**
- Debug log on both claim-failure paths (`Start skipped …` / `Retry ignored: …`).
- No-downloader branch marks Cancelled jobs Failed too — Start re-runs them, so leaving them Cancelled misreports why nothing happened. (This one had a directly testable failure; new test covers it.)
- `ClearAll` / `AbortAllActive` share one `CancelRunEpoch()`. Agreed this is the PR's core invariant and shouldn't have been duplicated.
- `"Batch complete"` no longer implies the queue is finished while a coalescing batch is still running.

**Not applied:** the per-job CAS flag instead of `HashSet` + lock. Contention here is two Start clicks and a Retry; the lock is uncontended and the set is the clearer statement of "which jobs are claimed". Happy to switch if you'd rather.

**Thanks for the two corrections** — the flakiness call on the first test and the "Start resurrects tile-cancelled jobs" regression flag were both wrong, and you refuted them before they reached me. The severity context is fair too: `AsyncRelayCommand` disabling the Start button meant the reachable route was the Retry/Start interleave, not pure Start-clicking.

New tests: `RetryJobAsync_DoesNotStartAJobAlreadyWaitingForAWorkerSlot`, `StartAllAsync_WithoutADownloader_FailsCancelledJobsToo`.
Full suite: **4869 passed, 0 failed, 2 skipped**. `DiffusionNexus.sln` builds clean.

PR #547, opened 2026-08-28
Civitai API gateway: stop the browser and downloads tripping 429

Little-God1983:
## The problem

Downloading from the LoRA Viewer / Civitai Browser trips HTTP 429. The cause is structural rather than a single bug: **17 production call sites share one `ICivitaiClient`, and only 4 of them — all inside the library-sync pipeline — went through the request pacer.** Everything else fired unpaced and concurrently, each surface discovered a rate limit on its own while the others kept digging, and the same model page was fetched repeatedly within seconds.

Most notably, **every download fetched `GET models/{id}` twice**: once in `LoraDownloadService` to persist metadata, then again seconds later in the completion sync just to read tags.

## What this does

Adds `CivitaiApiGateway`, a decorator over `ICivitaiClient` registered **as** `ICivitaiClient`, so all 17 call sites get pacing, caching and 429-awareness without being edited. It composes three small collaborators, all shared singletons:

- **Two-lane pacing off one timestamp** — interactive 750 ms (browser, detail panel, dialogs, downloads), background 1500 ms (library sync, the visible-tile update sweep). One shared timestamp is what makes background work space itself *behind* the user instead of alongside them.
- **A shared adaptive 429 cooldown** — the raw client now reports every 429 the moment it sees one (including ones its own retry recovers from), so one rate limit pauses every surface instead of only the caller that drew it. Spacing widens up to 4× and decays after five quiet minutes. `Retry-After` is now parsed in its HTTP-date form too, which was previously read and silently discarded.
- **A bounded single-flight TTL cache** — model/version 15 min, by-hash 60 min, search 2 min, 1000 entries. This is what collapses the duplicate call per download, the detail panel's refetch, and the applier's repeat for multi-version models.

Plus five call-site fixes the decorator cannot make for itself:

| | Change | Saving |
|---|---|---|
| F1 | Read the `.civitai.info` sidecar **before** the by-hash lookup in sync's identify step (the Sorter always did it in this order) | 2 API calls **and** a multi-gigabyte hash per sidecar-bearing file |
| F2 | One `models/{id}` call for a model's images instead of one `model-versions/{id}` per version, with a per-version fallback for versions the page omits | M→1 per multi-version model |
| F3 | The browser no longer searches from its constructor — opening the LoRA Viewer used to spend up to 10 paginated searches on a tab you may never visit. Sort/period/type changes are now debounced like the text box | up to 10 calls per viewer open |
| F4 | The download queue's restore-time version rehydrate now sends the API key | authenticated quota; gated versions visible |
| F5 | Six sites building an `HttpClient` per operation now share one, with `PooledConnectionLifetime` set | connection reuse |

## Notable design points

- `CivitaiRateLimitedException` derives from `HttpRequestException` with `StatusCode = TooManyRequests`, so every existing `catch (…) when (ex.StatusCode == TooManyRequests)` handler keeps working untouched.
- The in-call 429 retry budget drops 3 → 1. Three retries meant a single call could sleep 10 + 20 + 40 s *inside* the call while holding a download slot; the shared cooldown does that waiting properly now. The 5xx budget stays at 3.
- Pacing moved from four hand-placed call sites into one gateway; `LoraUpdateChecker`'s private 60-second backoff is deleted as superseded.
- `DiffusionNexus.Civitai` keeps its zero project references and zero NuGet packages.

## Known limitation

The two lanes equalise the request *interval* but grant no *queue* priority — the pacer is a single `SemaphoreSlim(1,1)` served FIFO, so an interactive call can still queue behind roughly 8 background requests. Bounded, far better than the current 429 storms, and documented in the design spec; a two-queue pacer is the follow-up.

## Test plan

- Full suite: **4925 passed / 0 failed / 2 skipped** (the 2 are opt-in online canaries). Release build: 0 errors, no new warnings.
- `new CivitaiClient(` appears exactly once in production code — the DI registration. No call site bypasses the gateway.
- **Manual smoke still owed:** queue 5+ downloads while a library sync runs and watch the Unified Console (Network) for 429s, confirming one `models/{id}` per download instead of two; and confirm Browse Civitai populates on first tab open (its `OnAttachedToVisualTree` hook has no `DataContextChanged` fallback and no unit test can cover that).

Design: `docs/superpowers/specs/2026-08-28-civitai-api-gateway-design.md`


DiffusionNexus.Civitai/CivitaiResponseCache.cs
```diff
@@ -0,0 +1,214 @@
...
+        // genuine race, and every caller — winner and losers alike — awaits the same task.
+        var lazy = _inFlight.GetOrAdd(key, _ => new Lazy<Task<object?>>(
+            () => RunAsync(key, ttl, factory, generation), LazyThreadSafetyMode.ExecutionAndPublication));
+        try
+        {
+            return (T?)await lazy.Value.WaitAsync(ct).ConfigureAwait(false);
+        }
+        finally
+        {
+            _inFlight.TryRemove(new KeyValuePair<string, Lazy<Task<object?>>>(key, lazy));
```
Little-God1983: **A caller that cancels — or the leader itself — evicts the shared in-flight entry, so single-flight breaks exactly in the cancel-heavy path this PR targets.**

`finally` runs for *every* awaiter, including one whose `WaitAsync(ct)` threw. When a joiner (or the leader) cancels while the fetch is still running, its `finally` removes the `_inFlight` entry even though the shared `Lazy` task is still alive. The next caller for the same key then misses `TryGet` (nothing has been written yet) and starts a **second** concurrent request for the same key.

Concrete: the update checker starts `model:123`; the detail panel joins; the update checker abandons its page on a pagination/filter change → its `finally` drops the entry → the next visible-tile sweep hits `model:123` again and issues a duplicate GET while the first is still in flight.

`LeaderCancellation_DoesNotCancelAJoinerWaitingOnTheSameFetch` covers the joiner *surviving*, but nothing covers the entry being removed out from under a live fetch. Removal should be owned by the fetch, not by each awaiter — e.g. do it in `RunAsync`'s own `finally`, or hang it off `lazy.Value` with a continuation.

DiffusionNexus.Civitai/CivitaiClient.cs
```diff
@@ -213,6 +219,25 @@ private TimeSpan TransientDelay(int attempt)
...
+    /// </summary>
+    private static TimeSpan? ParseRetryAfter(HttpResponseMessage response)
+    {
+        var header = response.Headers.RetryAfter;
+        if (header is null) return null;
+        if (header.Delta is { } delta) return delta;
+        if (header.Date is { } date)
+        {
+            var wait = date - DateTimeOffset.UtcNow;
+            return wait > TimeSpan.Zero ? wait : TimeSpan.Zero;
```
Little-God1983: **`Retry-After` is never clamped, and both consumers feed it straight into `Task.Delay`.**

`ParseRetryAfter` clamps a *past* date to zero but puts no ceiling on a future one. The value flows to two unbounded sleeps:

* `CivitaiClient.cs:300` — `await Task.Delay(RateLimitDelay(attempt, serverRetryAfter), ct)`
* `CivitaiRateLimitCooldown.WaitAsync` → `_delay(remaining, ct)` (`CivitaiRateLimitCooldown.cs:99`), reached via `CivitaiApiGateway.SendAsync` on `CancellationToken.None`

`Task.Delay` throws `ArgumentOutOfRangeException` for anything over ~49.7 days. `date - DateTimeOffset.UtcNow` uses the **local wall clock**, so a machine whose clock is years behind (or a server that emits a bad date) yields a multi-year `TimeSpan` → `ArgumentOutOfRangeException` out of the gateway, which no caller expects: `SyncFaults.IsItemFault` won't match it, so it escapes the item loop and aborts the whole sync run.

Even short of the throw, an unremarkable `Retry-After: 7200` freezes **every** Civitai call in the process for two hours with no cap — `MaxIntervalMultiplier` bounds the spacing but nothing bounds the cooldown deadline. Clamp here (e.g. `Math.Min(wait, MaxRetryAfter)` with a few-minutes ceiling) so both consumers inherit the bound.

DiffusionNexus.Service/Services/Sync/Steps/IdentifyModelStep.cs
```diff
@@ -198,7 +193,16 @@ public async Task<SyncItemResult> ExecuteOneAsync(SyncItem item, string? apiKey,
                 return SyncItemResult.Success;
             }
 
-            // No Civitai match, no sidecar — read the file's own header, then guess from its name.
+            var hash = await ResolveHashAsync(uow, candidate, ct).ConfigureAwait(false);
```
Little-God1983: **Sidecar-first means a file with a sidecar never gets its real bytes hashed, and the sidecar''s *claimed* SHA256 becomes the stored one.**

`ResolveHashAsync` is now only reached when the sidecar applied nothing. For every sidecar-bearing file the step returns at line ~192 before hashing, so `ModelFile.HashSHA256` is filled instead by `SidecarMetadataApplier`:

```csharp
if (hashes.TryGetProperty("SHA256", out var sha256) && sha256.GetString() is { } sha256Str)
{ dbFile.HashSHA256 ??= FileHasher.NormalizeSha256(sha256Str); dirty = true; }
```

That inverts the invariant the old ordering existed to hold — the comment deleted from `ResolveHashAsync`'s call site said it outright: *"Set before the Civitai applier runs: it fills hashes with `??=`, so the digest of the bytes we actually have must already be in place to win over the response's (possibly stale) one."* The sidecar applier has the same `??=`, and it now runs first.

Consequences: a truncated / partially-downloaded / hand-swapped file records the hash Civitai *says* it should have; `LoraDuplicateFinder` and the startup hash-repair pass then both trust a value nothing verified. And because `ResolveHashAsync` short-circuits on `FileHasher.IsSha256(candidate.Sha256)`, the real digest is never computed on any later run either.

DiffusionNexus.Civitai/CivitaiApiGateway.cs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /// <inheritdoc />
+    public Task<CivitaiModel?> GetModelAsync(int modelId, string? apiKey = null, CancellationToken cancellationToken = default)
+        => SendAsync(CivitaiResponseCache.ModelKey(modelId), ModelTtl, apiKey,
+            ct => _inner.GetModelAsync(modelId, apiKey, ct), cancellationToken);
+
+    /// <inheritdoc />
+    public Task<CivitaiModelVersion?> GetModelVersionAsync(int modelVersionId, string? apiKey = null, CancellationToken cancellationToken = default)
+        => SendAsync(CivitaiResponseCache.VersionKey(modelVersionId), VersionTtl, apiKey,
```
Little-God1983: **A 15-minute version cache silently defeats the waitlist''s user-initiated re-check, and the invalidation hooks that would fix it are never called.**

`CivitaiWaitlist.RefreshEntryAsync` exists to poll for an early-access deadline expiring; it goes through this method. Two "Re-check" presses inside 15 minutes now return the byte-identical cached `CivitaiModelVersion`, so `IsEarlyAccessActive()` re-evaluates against a stale `earlyAccessDeadline` / `paidAccess` and the button appears to do nothing — the entry stays `Waiting` past the moment it went free. Same for `ModelDetailViewModel`'s refresh and the download queue's rehydrate.

`ICivitaiApiCache.InvalidateVersion` / `InvalidateModel` are exactly the tool for this, but `grep` finds **no production caller** for either — the only cache calls anywhere are `Clear()` (token dialog + `LibrarySyncService`). Either wire `InvalidateVersion(entry.VersionId)` into the waitlist re-check (and `InvalidateModel` into the detail-panel refresh), or the two methods are dead API surface.

DiffusionNexus.UI/Views/Dialogs/CivitaiTokenDialog.axaml.cs
```diff
@@ -98,33 +100,43 @@ private void OnCivitaiLinkClick(object? sender, PointerPressedEventArgs e)
...
-                return (true, null);
-
-            if (response.StatusCode is System.Net.HttpStatusCode.Unauthorized
-                or System.Net.HttpStatusCode.Forbidden)
-            {
-                return (false, "Invalid API token. The token was rejected by Civitai (401/403).");
-            }
-
-            return (false, $"Civitai returned an unexpected status: {(int)response.StatusCode} {response.ReasonPhrase}");
+            await client.GetModelsAsync(new CivitaiModelsQuery { Limit = 1 }, apiKey: token);
```
Little-God1983: **Token validation lost its 15-second timeout and gained an unbounded, uncancellable wait.**

The removed raw client had `Timeout = TimeSpan.FromSeconds(15)`. The gateway path has no per-call bound at all:

* `CivitaiApiGateway.SendAsync` runs the cooldown + pacer waits on `CancellationToken.None` — during an active 429 cooldown that is `Retry-After` (unclamped, see `ParseRetryAfter`) or the 30 s default;
* `CivitaiClient`'s own 429 retry adds another `Task.Delay(Retry-After)`;
* the shared `HttpClient` in `AddCivitaiGateway` is constructed with no `Timeout`, so it falls back to the 100 s default rather than 15 s;
* no `CancellationToken` is passed here, so the dialog cannot abandon its own wait either.

Result: a user who opens the token dialog while a sync has just drawn a 429 sits on "Validating…" for minutes with no cancel and no timeout message. Pass a `CancellationTokenSource(TimeSpan.FromSeconds(15))` token to `GetModelsAsync` so the existing `catch (TaskCanceledException)` branch still means what it says.

DiffusionNexus.Civitai/CivitaiApiGateway.cs
```diff
@@ -0,0 +1,201 @@
...
+        Func<CancellationToken, Task<T?>> call, CancellationToken ct)
+        where T : class
+    {
+        // Delegated to the cache rather than tracked here: _cache is the ONE object shared by
+        // both lane instances (interactive and background wrap the same cache), so the last-seen
+        // key has to live where it is actually shared, not on a per-lane field that would leave
+        // each lane with its own, inconsistent memory of a single store.
+        _cache.NoteApiKey(apiKey);
+
+        return await _cache.GetOrAddAsync(cacheKey, ttl, async () =>
```
Little-God1983: **Detached fetches turn routine 429s and network blips into `UNOBSERVED TASK EXCEPTION` error-log entries.**

Running the factory on `CancellationToken.None` means the shared task outlives every caller. When all callers have abandoned via `WaitAsync(ct)` and the fetch *then* faults — a `CivitaiRateLimitedException`, a DNS blip, a `JsonException` — nothing observes `lazy.Value`, so the fault reaches `App.axaml.cs:85`:

```csharp
TaskScheduler.UnobservedTaskException += (sender, args) =>
{
    Serilog.Log.Error(args.Exception, "UNOBSERVED TASK EXCEPTION");
    FileLogger.LogError($"UNOBSERVED TASK EXCEPTION: {args.Exception?.Message}", args.Exception);
    args.SetObserved();
};
```

Every abandoned-then-failed gateway call is logged as an ERROR with a label that reads like a bug, at GC time and detached from the operation that caused it. The XML doc weighs the cost of the detached fetch but not this. Attach a no-op observer to the task (`_ = lazy.Value.ContinueWith(t => _ = t.Exception, TaskContinuationOptions.OnlyOnFaulted)`) so an abandoned failure stays a quiet cache miss.

DiffusionNexus.Service/Services/Sync/LibrarySyncService.cs
```diff
@@ -137,6 +144,21 @@ public async Task<SyncReport> ExecuteAsync(SyncPlan plan, IProgress<LibrarySyncP
...
+            // Viewer's per-tile "Download Metadata", which forces identify and thumbnails.
+            //
+            // Inside the try (not between _gate.Wait(0) and it): _apiCache.Clear() realistically
+            // never throws, but if it ever did, throwing between the gate acquire and the try
+            // would skip the finally below and leak the gate — wedging sync for the rest of the
+            // session with no way to release it short of restarting the app.
+            var options = plan.Options;
+            if (options.ForceIdentify || options.ForceTags || options.ForceImages || options.ForceThumbnails)
+            {
+                _apiCache?.Clear();
```
Little-God1983: **One tile''s "Download Metadata" wipes the whole process-wide response cache.**

The condition fires on *any* force flag, and the comment acknowledges it covers the LoRA Viewer's per-tile "Download Metadata" — which forces identify + thumbnails for a **single** model. `Clear()` also bumps the generation, so every fetch already in flight for every other surface (the browser's current page, the detail panel, the update sweep) is stamped stale and its result thrown away.

A user clicking "Download Metadata" down a list of ten tiles therefore empties the shared cache ten times, which is precisely the repeat-request pattern the cache was added to absorb. `ICivitaiApiCache.InvalidateModel(civitaiModelId)` is the right-sized tool and currently has no caller — scope the drop to the models the plan actually targets, and keep `Clear()` for a full-library forced re-sync.

DiffusionNexus.Service/Services/Sync/Steps/FetchImagesStep.cs
```diff
@@ -98,14 +102,26 @@ public async Task<SyncItemResult> ExecuteOneAsync(SyncItem item, string? apiKey,
...
+                    apiKey,
+                    ct)
+                .ConfigureAwait(false);
+
             foreach (var candidate in candidates)
             {
                 ct.ThrowIfCancellationRequested();
 
                 // Which version the refusal handler names in its one warning line.
                 inFlight = candidate;
```
Little-God1983: **A refusal raised by the new model-page call is reported against a version id it has nothing to do with.**

`ApplyImagesFromModelAsync` (line 107) now issues `GET models/{civitaiModelId}` *before* the loop, but `inFlight` is only assigned inside the loop — so when that model-page request is refused (401/403/early-access), `inFlight` is still `candidates[0]` and `RecordRefusalAsync` logs:

> `{item.Name}: Civitai refused ({code}) for id {inFlight.CivitaiVersionId}; marked as checked`

naming a *version* id for a refusal that came from the *model* endpoint. Anyone debugging from the log will go look up a version Civitai never rejected. Either set `inFlight` before the batched call or give `RecordRefusalAsync` a way to say "the model page".

DiffusionNexus.Civitai/CivitaiApiGateway.cs
```diff
@@ -0,0 +1,201 @@
...
+
+    // Not cached: unused in production, and an images/tags/creators listing is a browse of a
+    // moving target rather than a record we would want to hold on to.
+
+    /// <inheritdoc />
+    public async Task<CivitaiPagedResponse<CivitaiModelImage>> GetImagesAsync(
+        CivitaiImagesQuery? query = null, string? apiKey = null, CancellationToken cancellationToken = default)
+    {
+        _cache.NoteApiKey(apiKey);
+        await _cooldown.WaitAsync(cancellationToken).ConfigureAwait(false);
```
Little-God1983: **Two different cancellation contracts inside one class, plus three copies of the same preamble.**

`GetImagesAsync` / `GetTagsAsync` / `GetCreatorsAsync` each repeat `await _cooldown.WaitAsync(ct); await _pacer.WaitAsync(Interval, ct);` and pass the **caller''s** token, while `SendAsync` deliberately runs the identical pair on `CancellationToken.None`. Same class, same two waits, opposite answers to "can a caller''s cancellation tear down a rate-limit wait?" — and the long `SendAsync` remark explaining why detaching is required doesn''t apply here only because these three aren''t single-flighted, which is not obvious at the call site.

Worth factoring the preamble into one private `await GateAsync(ct)` and stating the rule once: shared/single-flighted work detaches, exclusive work honours the caller.

DiffusionNexus.Civitai/CivitaiRateLimitCooldown.cs
```diff
@@ -0,0 +1,102 @@
...
+    public void OnRateLimited(TimeSpan? retryAfter)
+    {
+        lock (_lock)
+        {
+            var now = _clock();
+            var wait = retryAfter ?? DefaultCooldown;
+
+            // Extend, never shorten: a second 429 while a longer cooldown is running must not
+            // release everyone early.
+            var deadline = now + (long)wait.TotalMilliseconds;
```
Little-God1983: **The multiplier doubles once per 429 *report*, and the client now reports twice per rate-limited call — so the first rate limit a user hits jumps straight to the 4× cap.**

`CivitaiClient.GetAsync` calls `_rateLimitObserver.OnRateLimited(...)` before deciding whether to retry, and `maxRateLimitRetries = 1` means a call that is refused twice reports twice: `_multiplier` goes 1 → 2 → 4 inside a single request. Concurrent callers discovering the same limit multiply it further (capped, but instantly).

Effect: one rate-limit event widens interactive spacing from 750 ms to 3 s and background from 1.5 s to 6 s for the full 5-minute `MultiplierDecay`, which reads as "the app got slow for no reason" rather than the graduated back-off the doc describes. Consider debouncing the escalation (ignore reports within, say, the current cooldown window) so the multiplier tracks *episodes* rather than individual 429 responses.

DiffusionNexus.Service/Services/Sync/SyncServiceCollectionExtensions.cs
```diff
@@ -51,7 +52,16 @@ public static IServiceCollection AddLibrarySync(this IServiceCollection services
         // ids the tags and images steps need, and only a fetched image record has a URL to make a
         // thumbnail from.
         services.AddTransient<ISyncStep, DiscoverFilesStep>();
-        services.AddTransient<ISyncStep, IdentifyModelStep>();
+
+        // The only step holding a client of its own — the by-hash lookup. Background lane, and
```
Little-God1983: **Convention: the refactored pacer types were deleted outright rather than marked obsolete.**

`.github/copilot-instructions.md` states:

> "When a refactoring task is issued then mark the method/function/class you are refactoring as obsolete and to be removed later to avoid code duplication/dead code."

This PR deletes `DiffusionNexus.Service/Services/Sync/CivitaiRequestPacer.cs` wholesale — the public `ICivitaiRequestPacer`, `CivitaiRequestPacer` and `NoCivitaiRequestPacer` — and removes the `pacer` parameters from `CivitaiMetadataApplier` and `IdentifyModelStep`, all in one step. `DiffusionNexus.Service` is a public-surface project, so anything outside the solution binding to those types or ctor signatures breaks with no deprecation window. An `[Obsolete("Moved to DiffusionNexus.Civitai; use CivitaiApiGateway")]` shim forwarding to the new type would satisfy the rule and keep the ctor overloads compiling.

DiffusionNexus.UI/ViewModels/CivitaiBrowser/CivitaiBrowserViewModel.cs
```diff
@@ -521,9 +535,12 @@ private async Task DebouncedSearchAsync()
     #region Property change hooks
 
     partial void OnSearchTextChanged(string value) { if (_initialized) _ = DebouncedSearchAsync(); }
-    partial void OnSelectedSortChanged(string? value) { if (_initialized) _ = SearchAsync(); }
-    partial void OnSelectedPeriodChanged(CivitaiPeriod value) { if (_initialized) _ = SearchAsync(); }
-    partial void OnSelectedModelTypeChanged(ModelTypeOption? value) { if (_initialized) _ = SearchAsync(); }
+
+    // Debounced like the text box: a user comparing two sorts or three periods used to spend a
+    // full paginated search on each intermediate choice.
+    partial void OnSelectedSortChanged(string? value) { if (_initialized) _ = DebouncedSearchAsync(); }
```
Little-God1983: **Debouncing sort/period/type opens a 400 ms window in which `_nextCursor` still belongs to the previous query.**

These three used to call `SearchAsync()` synchronously, and `SearchAsync` resets `_nextCursor = null` (line 243). Now the reset is deferred by `Task.Delay(400)`, so for 400 ms after the user changes Sort / Period / Model type the VM still advertises `HasMore` from the old query and `LoadMoreAsync` (line 258) will paginate the **old** cursor under the new filter.

That is reachable without a fast click: `ApplyClientSideFilters` -> `MaybeAutoLoadMoreAsync` (line ~952) fires a load-more automatically whenever the visible count drops below `AutoLoadMoreThreshold`, which is exactly what a model-type change does. The result is a grid mixing pages from two different queries, with the newly selected sort applied to neither.

Clearing `_nextCursor` (and raising `HasMore`) in the property hook, before the debounce, keeps the debounce without the window.

DiffusionNexus.Civitai/CivitaiResponseCache.cs
Little-God1983: Confirmed and fixed in 90525b64. Removal is now owned by the fetch — it happens in `RunAsync`'s own `finally` rather than in each awaiter's, so a cancelled caller can no longer evict a live entry. The atomic key+value `TryRemove` is kept so a newer entry for the same key is never removed. New test `ACancelledAwaiter_DoesNotEvictALiveInFlightEntry` asserts one inner call where the old code made two.

Follow-up in a48b028c: per-key invalidation had the mirror-image hole — it dropped `_entries` but not the in-flight write-back, so a user-pressed re-check could still be answered with pre-invalidation data. There is now a per-key generation counter alongside the global one; `Invalidate` deliberately leaves `_inFlight` alone (so no duplicate request starts) and only suppresses the stale write-back.

DiffusionNexus.Civitai/CivitaiClient.cs
Little-God1983: Confirmed and fixed in ae5c20c8. `ParseRetryAfter` now clamps to a 10-minute ceiling on both the delta and the HTTP-date branch, so the single parse site bounds both consumers — the client's own retry delay and the cooldown deadline. The past-date clamp is unchanged.

10 minutes because real Civitai values are low tens of seconds, so it is generous headroom for a genuine instruction while keeping a bad server date or a skewed client clock to minutes of degraded service rather than an `ArgumentOutOfRangeException` that would escape `SyncFaults.IsItemFault` and abort a whole sync run.

DiffusionNexus.Service/Services/Sync/Steps/IdentifyModelStep.cs
Little-God1983: This was the most serious of the twelve and it is fixed in d552e398. You are right about the mechanism and about the consequence.

`ResolveHashAsync` is now called inside the `sidecar.Applied` branch before the stamp, so the measured digest is committed with a plain assignment that overwrites whatever `SidecarMetadataApplier` put there with `??=`. A mismatch between the sidecar's claim and the real bytes is logged as a warning. The zero-Civitai-calls saving that motivated the reorder is preserved — nothing in that branch touches the network.

The cost is that a sidecar-bearing file is hashed again, which the reorder had been saving. Correctness over speed: the branch exists to cut API calls, not hashing. Test added with a deliberately wrong sidecar SHA256, verified failing before the fix. Stale claims in the plan doc and two in-body comments corrected in 444645fc.

DiffusionNexus.Civitai/CivitaiApiGateway.cs
Little-God1983: Confirmed and fixed in 8a1fe094. `InvalidateVersion(entry.VersionId)` now runs before the waitlist re-check, and `InvalidateModel` before the detail panel's fetch. Scoped deliberately to user-pressed paths: the waitlist's 1-minute timer calls only the API-free availability recompute, and the download queue's rehydrate-on-resume is left cached because it fires once per job rather than as a repeated re-check.

Both methods now have production callers, so neither is dead surface any more. `InvalidateModel` also gained one from the fix to your `LibrarySyncService` comment.

DiffusionNexus.UI/Views/Dialogs/CivitaiTokenDialog.axaml.cs
Little-God1983: You are right, and so was your reasoning about it being both unbounded and uncancellable — I had waved this one through earlier on the grounds that it was a user-initiated action. Fixed in 16030d69: a `CancellationTokenSource(15s)` token is passed into the gateway call, restoring the old bound, and the `catch (TaskCanceledException)` is narrowed with `when (cts.IsCancellationRequested)` so it means what it says.

DiffusionNexus.Civitai/CivitaiApiGateway.cs
Little-God1983: Confirmed and fixed in 90525b64, folded into the in-flight lifetime fix rather than adding a second mechanism: a fault observer is attached to the shared task inside the `Lazy` factory, so it is in place before any awaiter can receive the task and it also covers an already-faulted one.

Related: the test I first added for this turned out to be vacuous — it still passed with the observer removed, because it depended on GC timing. Deleted in a48b028c rather than left as false assurance.

DiffusionNexus.Service/Services/Sync/LibrarySyncService.cs
Little-God1983: Confirmed and fixed in d9818862. A model-scoped forced run now invalidates per model instead of calling `Clear()`; `Clear()` is kept for a genuinely library-wide or folder-wide forced re-sync. The target set comes from `SyncScope.ModelIds`, which the per-tile path already populates via `SyncScope.ForModels(modelId)`.

Two follow-ups in 02daef49, because the first cut under-reached: a forced identify also goes through `GetModelVersionByHashAsync`, cached under `hash:{HASH}` for 60 minutes, which `InvalidateModel` did not touch — so "force" would still have been answerable from cache for an hour. `ICivitaiApiCache` gained `InvalidateHash`, and the scoped path now drops the model page, every version, and every file hash the targeted models own. The call also moved inside the inner `try`, since it is now awaited DB I/O and had been sitting outside the block that makes `ExecuteAsync` total, and it falls back to `CivitaiId` when `CivitaiModelPageId` is null so legacy rows are not silently skipped.

DiffusionNexus.Service/Services/Sync/Steps/FetchImagesStep.cs
Little-God1983: Confirmed and fixed in c1431e69. `inFlight` is nullable now and stays null across the model-page call, so the refusal log distinguishes "the model page" from a specific version id rather than naming a version Civitai never rejected.

DiffusionNexus.Civitai/CivitaiApiGateway.cs
Little-God1983: Fair, and fixed in d08eef9c. The preamble is now a single private `GateAsync` helper with four call sites, and the rule is stated once in its remark: shared, single-flighted work detaches from any one caller's token; exclusive work honours the caller. Behaviour is unchanged in both directions — `SendAsync` still passes `CancellationToken.None`, the three uncached methods still pass the caller's token.

DiffusionNexus.Civitai/CivitaiRateLimitCooldown.cs
Little-God1983: Confirmed and fixed in 3b9fd73e. The escalation is now gated on whether a report arrives while a cooldown from a previous report is still active, so the multiplier tracks episodes rather than individual 429 responses — a single rate-limited call that reports twice now leaves it at 2 rather than jumping to the 4x cap. Extend-never-shorten is untouched, and a genuinely new episode after the decay window still escalates.

One residual you may want to know about: when `Retry-After` is zero the deadline equals now, so two reports in one call can still double twice. Narrow, and the past-date clamp is the only way to reach it.

DiffusionNexus.UI/ViewModels/CivitaiBrowser/CivitaiBrowserViewModel.cs
Little-God1983: Confirmed and fixed in ea2bbe3c. `_nextCursor` is cleared and `HasMore` raised synchronously in the property hook, before the debounce, so the 400 ms window is closed while the search itself stays debounced. Both readers of the stale cursor are covered — `LoadMoreAsync` and the auto-top-up path both guard on `HasMore`, which derives purely from `_nextCursor`.

The `_initialized` guard and the client-side-only toggles are untouched. Tests added and verified failing against the unfixed code.

DiffusionNexus.Service/Services/Sync/SyncServiceCollectionExtensions.cs
Little-God1983: Pushing back on this one rather than actioning it, and flagging it for @Little-God1983 to overrule if you disagree.

The convention is real and I should have weighed it — `.github/copilot-instructions.md` does say to mark a refactored type obsolete rather than delete it. But the same sentence gives its purpose: *"to avoid code duplication/dead code"*. Adding `[Obsolete]` forwarding shims here would produce exactly both.

The specifics:

- The types did not disappear, they **moved within this solution** — `ICivitaiRequestPacer`/`CivitaiRequestPacer` are now in `DiffusionNexus.Civitai`, because that project is where the gateway lives and it has zero project references by design (the pacer could not stay in `Service` without `Civitai` taking a dependency on it).
- Every consumer was updated in the same commit, and the solution builds green. A shim would be referenced by nothing.
- The premise that `DiffusionNexus.Service` is a public surface with external binders does not hold for this repo: it has no `IsPackable`/`PackageId`/`GeneratePackageOnBuild`, and `grep` for `DiffusionNexus.Service.csproj` finds exactly three consumers, all inside this solution — `DiffusionNexus.UI`, `DiffusionNexus.Tests`, `DiffusionNexus.IntegrationTests`. Nothing outside the solution can bind to those types or ctor signatures, so there is no deprecation window to preserve. (The published-package repo is the separate Installer SDK.)
- `NoCivitaiRequestPacer` specifically was deleted because it had become genuinely unreferenced once both consumers dropped their optional-pacer fallback — a review earlier in this branch flagged it as dead code whose only remaining use would be to let a future caller opt out of the shared pacer, which is the thing this PR exists to prevent.

So following the letter here would reintroduce the dead code the rule is written to avoid. Happy to add the shims if you would rather have the convention applied uniformly — say the word and I will.

DiffusionNexus.Civitai/CivitaiRateLimitCooldown.cs
```diff
@@ -0,0 +1,121 @@
...
+    public void OnRateLimited(TimeSpan? retryAfter)
+    {
+        lock (_lock)
+        {
+            var now = _clock();
+            var wait = retryAfter ?? DefaultCooldown;
+
+            // Captured before _cooldownUntil is possibly extended below, so it reflects whether A
+            // COOLDOWN FROM AN EARLIER REPORT was still active when THIS report arrived.
+            var sameEpisode = _everRateLimited && now < _cooldownUntil;
```
Little-God1983: **Episode guard misses whenever the server sends `Retry-After`** — the 1 -> 2 -> 4 walk inside one request is still there.

`CivitaiClient.GetAsync` reports the 429 *before* deciding to retry, then sleeps `RateLimitDelay(attempt, serverRetryAfter)`, which in production is exactly `serverRetryAfter` (`RetryDelayOverride` is null outside tests).

So with `Retry-After: 5`:

1. report #1 -> `_cooldownUntil = now + 5000`, multiplier 2
2. client sleeps 5 s, retries, gets 429 again
3. report #2 arrives at `now + 5000 + ε` -> `now < _cooldownUntil` is **false** -> `sameEpisode` false -> multiplier 4

The first rate limit a user ever hits still jumps straight to the 4x cap — the exact scenario the commit message and the remark above describe. Only the *no*-header path is actually fixed (`DefaultCooldown` 30 s vs the 10 s exponential backoff).

`IntervalMultiplier_TwoReportsWithinTheSameCooldown_StaysAtTwo` passes only because its fake clock is never advanced between the two `OnRateLimited` calls; the real retry loop always advances it by the full `Retry-After`. Advancing `_now += 10_001` in that test reproduces the escalation.

A fix that survives the sleep would be an explicit episode marker (e.g. keep escalating off `_lastRateLimit` but require a gap larger than the last cooldown *plus* some slack, or have `GetAsync` report only the first 429 of a call).

DiffusionNexus.Civitai/CivitaiResponseCache.cs
```diff
@@ -0,0 +1,301 @@
...
+        {
+            var value = await factory().ConfigureAwait(false);
+
+            // A Clear() while this fetch was in flight (an API key change, an explicit reset) means
+            // the world moved on before this answer arrived. Writing it now would put a stale value
+            // back into a store that was just emptied specifically to get rid of stale values. A
+            // per-key InvalidateModel/InvalidateVersion/InvalidateHash for THIS key while the fetch
+            // was in flight is the same situation at a narrower scope — the key version check below
+            // catches that even when the global generation never moved.
+            if (Interlocked.Read(ref _generation) == generation && CurrentKeyVersion(key) == keyVersion)
```
Little-God1983: **Check-then-act: an `Invalidate*` landing between this check and the write on the next lines still gets its stale answer cached.**

Sequence:

1. fetch for `model:123` reaches this line, `_generation` and `CurrentKeyVersion(key)` both match -> enters the block
2. `InvalidateModel(123)` runs — `_entries.TryRemove` finds nothing (the entry does not exist yet), `_keyVersions` bumps to 1
3. the fetch executes `_entries[key] = new Entry(value, ...)`

The pre-invalidation page is now cached for the full `ModelTtl` (15 min; 60 min for a `hash:` key), which is precisely what `_keyVersions` was added to prevent — its own doc says "its answer is never written back into `_entries`".

Reachable from both new callers: `ModelDetailViewModel.FetchCivitaiDataAsync` invalidates on every detail-panel open, and `LibrarySyncService.InvalidateForcedScopeAsync` invalidates on every forced "Download Metadata", both of which routinely race a browser/update-sweep fetch for the same model id. The window is small but the consequence is a user-forced refresh silently answered from the page they were trying to correct.

Cheapest fix: stamp the captured `keyVersion` (and generation) into `Entry` and validate it in `TryGet`, so a late write is harmless rather than needing to be prevented.

DiffusionNexus.Service/Services/Sync/Steps/IdentifyModelStep.cs
```diff
@@ -353,23 +370,48 @@ private async Task<SyncItemResult> RecordMatchAsync(
...
+    /// execution touched the row — so this correctly re-hashes even when the sidecar branch (which
+    /// runs first and shares the same <paramref name="uow"/>) has, in the same call, already
+    /// written its own <c>??=</c> claim into the tracked entity: the plain assignment below
+    /// overwrites that claim with the measured digest regardless. Called both from the sidecar
+    /// path (to make sure a sidecar's claim never becomes the stored hash — see the class remarks)
+    /// and from the no-match path (where the hash also feeds the Civitai hash lookup itself).
+    /// </remarks>
+    private async Task<string> ResolveHashAsync(IUnitOfWork uow, IdentifyCandidate candidate, CancellationToken ct)
     {
         if (FileHasher.IsSha256(candidate.Sha256)) return candidate.Sha256!.ToUpperInvariant();
```
Little-God1983: **This short-circuit means the new "a sidecar's claim must never become the trusted hash" rule only applies to rows whose hash is still null.**

Before this change the sidecar branch ran first and `SidecarMetadataApplier` persisted the sidecar's *claimed* digest via `dbFile.HashSHA256 ??= FileHasher.NormalizeSha256(sha256Str)` (SidecarMetadataApplier.cs:640). Every row that path already touched now carries a well-formed SHA256, so `FileHasher.IsSha256(candidate.Sha256)` passes here and the file is returned verbatim — never measured.

Concretely: a stale or hand-edited `.civitai.info` sitting next to a re-downloaded or truncated file keeps supplying the wrong `ModelFile.HashSHA256` to the duplicate finder and to `GetModelVersionByHashAsync`, and a user-forced "Download Metadata" will not correct it. The new mismatch warning a few lines below can never fire for exactly the rows the bug already created.

The remark calls this "a stored, previously-verified digest", but a value written by the pre-fix sidecar path was never verified. If healing those rows is in scope, it needs either a one-off migration/backfill or a flag on the row recording whether the stored hash was measured or claimed.

DiffusionNexus.Civitai/CivitaiRateLimitCooldown.cs
Little-God1983: Confirmed and fixed in 0d4b177a. You are right that the previous guard only closed the no-header path — I verified `RateLimitDelay` returns `retryAfter` verbatim in production (`RetryDelayOverride` is a test-only seam), so the retry sleep equals the cooldown exactly and report #2 lands just past the deadline.

You also found something sharper than the finding states: the existing `IntervalMultiplier_DoublesPerRateLimitEpisode_CappedAtFour` was **encoding the bug**. Its timing — report, +1001 ms on a 1 s window, report again — is mechanically identical to the real single-call retry sequence, despite a comment claiming it represented separate episodes. It has been split into a same-call case (no escalation) and a genuinely-separate-episodes case (still escalates, identical timing), so the disambiguation is now what the tests actually exercise.

Fix: `OnRateLimited` takes `isRetryOfReportedCall`, set from a per-call flag in `GetAsync`. Chosen over "report only the first 429" because it keeps the deadline-extension guarantee — a longer `Retry-After` on the second report still extends the cooldown — for negligible extra complexity. New test drives the real sequence and asserts the multiplier ends at 2.

DiffusionNexus.Civitai/CivitaiResponseCache.cs
Little-God1983: Confirmed and fixed in 2a2967bc, and your suggested fix is the one taken: rather than trying to narrow the window, the captured generation and key version are now stamped into `Entry` and validated in `TryGet`, so a late write is harmless instead of needing to be prevented — a stale entry is ignored and dropped on read.

Worth recording that the race is a genuine thread race, not an async-ordering one: there is no `await` between the version check and the write, so reproducing it deterministically needed a small `internal` test hook at that point (precedented by `CivitaiClient.RetryDelayOverride`). The test drives the exact interleaving you described — invalidate after the check, before the write — and proves the value is not served.

DiffusionNexus.Service/Services/Sync/Steps/IdentifyModelStep.cs
Little-God1983: Right about the mechanism, and the comment is fixed in da21207b — it described the short-circuit's input as a previously-verified digest, which is untrue for any row the pre-fix sidecar path touched. It now states plainly what the short-circuit assumes and what it does not re-verify.

Scoping the rest deliberately, rather than building the migration/backfill or the measured-vs-claimed flag you offer as options. The sidecar-first ordering (`d9a1748c`) exists **only on this branch** — `git branch -a --contains` puts it on `feature/civitai-api-gateway` and nowhere else, never on develop or main. So no released build has ever written a claimed hash by that path, and the only databases that can carry a poisoned row are ones that ran this branch between `d9a1748c` and the fix in `d552e398`. In practice that is the maintainer's own machine during smoke testing.

A schema flag or a backfill migration for a defect that never shipped would be disproportionate, and the short-circuit earns its keep — it is what stops every sync re-hashing multi-gigabyte files. If the branch had merged before the fix I would answer differently. Flagged to the maintainer so they can re-run identification on any model whose hash looks wrong; happy to add a targeted repair if that turns out to matter.

PR #551, opened 2026-08-31
fix(viewer): route gated version downloads through the Browse tab's preflight

Little-God1983:
## Problem
In the LoRA viewer's **Installed** tab, opening a model's detail view and pressing **Download This Version** on an early-access or paywalled version ran the full token/destination/transfer flow and died with a bare red failure line. The **Browse Civitai** tab already handles this correctly: it badges gated versions and routes them through the preflight dialog with a waitlist option. The two surfaces disagreed, and the detail panel let the user walk into a guaranteed error.

## Fix — parity with the Browse tab
- **Badges**: version tabs now carry the browser version picker's exact classification (`IsEarlyAccessActive()` / `IsPermanentlyPaid()`; purple **EA** only for a gate that expires, red **PAID** wins otherwise) and render the same badges. A notice above the download button states the gating before any click.
- **Preflight**: the download command checks the gate *before* the token prompt and shows the existing `DownloadPreflightDialog` (which already hides "Download anyway" for paywalled versions and offers the waitlist only for expiring gates). Choice handling is an internal, unit-tested `ApplyGatedVersionChoice` — the single-version sibling of the browser's `ApplyPreflightChoice` (waitlist / open civitai.com or .red / cancel).
- **Shared waitlist**: `CivitaiWaitlist` gains a raw-data `TryAdd` overload (the browser-typed overload delegates to it), and the detail panel receives the Browse tab's waitlist **instance** — a second instance would clobber the persisted JSON snapshot. Entries added from the detail panel appear on the Waitlist tab and its badge immediately.

The 403 status-line fallback stays for anything detection misses.

## Tests
- `CivitaiVersionTabItemTests` (new): EA/PAID/free/expired classification on the tabs.
- `CivitaiWaitlistTests`: raw `TryAdd` payload capture, cross-overload dedup, permanent-paid rejection.
- `ModelDetailViewModelTests`: all `ApplyGatedVersionChoice` branches incl. the civitai.red host and cached-model-id fallback.
- Full suite: **5172 passed / 0 failed** (+ 11 integration tests green).

Manual GUI smoke still owed (needs a live gated model).


DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
```diff
@@ -411,6 +424,14 @@ private async Task DownloadSelectedVersionAsync()
         var tab = SelectedVersionTab;
         if (tab is null || tab.IsDownloaded) return;
 
+        // Gated versions can't be downloaded by the app at all — Civitai serves paid files
+        // through the website only. Mirror the Browse tab: one preflight dialog offering
+        // waitlist/website instead of letting the transfer die with a red failure line.
+        // Checked before the token prompt — no point demanding a key for a refused download.
+        if (tab.CivitaiVersion.IsEarlyAccessActive()
```
Little-God1983: **Correctness — the gate leaves entitled users no way to attempt the download.**

The preflight hides "Download anyway" whenever anything is paywalled (`DownloadPreflightDialog.ShowDownloadAnyway = !HasPaywalled`), so the `DownloadAnyway` branch of `ApplyGatedVersionChoice` is unreachable from the UI here — the only way to reach it is the headless owner-null fallback.

Scenario: a user whose Civitai key belongs to an account with the Supporter/membership entitlement — exactly the case `LoraDownloadService.cs:178` tells them to use (*"use a key from an account that has the entitlement"*) — selects an EA version and presses **Download This Version**. The dialog offers only Cancel / Skip Early Access / Open on Civitai / Add to waitlist; all four return `false`, so the transfer is never attempted. Before this change the request went out with the token and could succeed.

The repo contradicts itself on whether that is possible: `DownloadPreflightCopy.TemporaryTail` says Civitai *"only serves paid files through the website"*, while the `LoraDownloadService` 401/403 hint says an entitled key works. Whichever is right, one of the two should be corrected — and if the hint is right, this panel needs an explicit override path.

DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
```diff
@@ -493,6 +514,86 @@ private async Task DownloadSelectedVersionAsync()
...
+    /// <summary>
+    /// Shows the Browse tab's pre-download dialog for one early-access/paywalled version.
+    /// Needs a window owner; headless hosts (tests, tools) get <see cref="DownloadPreflightResult.DownloadAnyway"/>
+    /// so the plain download path — whose failure line still reports the 403 — stays reachable.
+    /// </summary>
+    private async Task<DownloadPreflightResult> ShowGatedVersionPreflightAsync(CivitaiVersionTabItem tab)
+    {
+        var owner = (Avalonia.Application.Current?.ApplicationLifetime
+            as Avalonia.Controls.ApplicationLifetimes.IClassicDesktopStyleApplicationLifetime)?.MainWindow;
+        if (owner is null) return DownloadPreflightResult.DownloadAnyway;
```
Little-God1983: **Fails open: a missing owner window removes the gate instead of blocking.**

If `Application.Current.ApplicationLifetime` is not an `IClassicDesktopStyleApplicationLifetime`, or `MainWindow` is null (window swap during startup/teardown), the gated version falls straight through to the token prompt and the transfer — producing exactly the bare red failure line this PR set out to remove, with no dialog, no log line and no status message explaining why.

Elsewhere the codebase treats a missing `MainWindow` as an error (`App.axaml.cs:932` throws). Failing *closed* — return `Cancel` and set a `StatusMessage` — would keep the fix intact in that path; the tests already call `ApplyGatedVersionChoice` directly, so they don't depend on the fail-open.

Secondary: `await dialog.ShowDialog(owner)` isn't wrapped in try/catch, so an owner in a bad state throws out of `DownloadSelectedVersionAsync` rather than degrading.

DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
```diff
@@ -493,6 +514,86 @@ private async Task DownloadSelectedVersionAsync()
...
+                return false;
+
+            case DownloadPreflightResult.OpenWebsite:
+                if (modelId > 0)
+                {
+                    // civitai.com hides NSFW from unauthenticated visitors; route those to the mirror.
+                    var host = isNsfw ? "civitai.red" : "civitai.com";
+                    OpenUrl($"https://{host}/models/{modelId}?modelVersionId={tab.CivitaiVersion.Id}");
+                }
+                return false;
```
Little-God1983: **Silent no-op when `modelId` resolves to 0.**

The guard has no `else`: if neither `tab.CivitaiVersion.ModelId` nor `CachedCivitaiModel?.Id` yields an id (older cached sidecar payload that omits `modelId`, or tabs built while `CachedCivitaiModel` is null), the user clicks **Open on Civitai**, the dialog closes, no browser opens, and nothing is logged or shown. The download just looks broken.

Every other branch of `ApplyGatedVersionChoice` reports something — a `StatusMessage` or a log line. This one should too:

```suggestion
            case DownloadPreflightResult.OpenWebsite:
                if (modelId > 0)
                {
                    // civitai.com hides NSFW from unauthenticated visitors; route those to the mirror.
                    var host = isNsfw ? "civitai.red" : "civitai.com";
                    OpenUrl($"https://{host}/models/{modelId}?modelVersionId={tab.CivitaiVersion.Id}");
                }
                else
                {
                    StatusMessage = $"No Civitai model id for '{ModelName}' — run 'Download Metadata' first.";
                    _logger?.Warn(LogCategory.Download, "LoraDownload",
                        $"Cannot open Civitai page: no model id for '{ModelName} — {tab.Label}'.");
                }
                return false;
```

DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
Little-God1983: @
Partially applied in 16e8b249. The copy contradiction is real and the older text lost: the LoraDownloadService hint now matches the decided stance instead of advising an entitled key. That stance is d0e7bfb4 ("drop the download-anyway escape hatch for paywalled versions", 2026-08-13, one day after the hint was last touched): the key-appended retry is refused either way, so the Browse tab deliberately ships without an override — and this PR exists to match the Browse tab. Resurrecting the escape hatch would undo that decision for both tabs; if the entitled-key claim is ever verified to work in-app, that is a product-level revisit of d0e7bfb4, not a patch to this panel.
@

DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
Little-God1983: @
Fixed in 16e8b249 — the gate now fails closed: a missing owner window yields Cancel plus a status line ("...paywalled on Civitai and the options dialog could not be shown — download not attempted") and a warn log, and ShowDialog is wrapped so a throwing owner degrades to Cancel instead of escaping DownloadSelectedVersionAsync. Covered by GatedDownload_WithNoOwnerWindow_FailsClosedInsteadOfBypassingTheGate.
@

DiffusionNexus.UI/ViewModels/ModelDetailViewModel.cs
Little-God1983: @
Fixed in 16e8b249 as suggested — the modelId==0 path now sets a status line pointing at "Download Metadata" and warn-logs, so the branch reports like every other one. Covered by GatedChoice_OpenWebsite_WithNoModelId_ReportsInsteadOfSilentlyDoingNothing.
@

Little-God1983:
**Follow-up in a8043e64 — third surface gated.** The toolbar "Download Lora" URL dialog had the same hole: pasting a link to a gated version ran the transfer into the 403. Now the dialog preview shows the same EA/PAID notice the moment the search resolves a gated version, and the download path routes through the preflight (waitlist / website / cancel). The preflight logic was extracted into a shared `GatedDownloadPreflight` — the detail panel delegates to it, so there are no longer two (about to be three) copies of the choice handling; the detail-panel tests survived the extraction unchanged. `DownloadLoraResult` now carries ModelId/Category/IsNsfw so waitlist entries from this surface are complete. Suite: 5180 unit + 11 integration, all green.

PR #552, opened 2026-08-31
feat(ui): searchable base-model picker for detail card + training runs

Little-God1983:
## What

The base-model dropdowns on the **model detail card**, the **Create Training Run dialog**, and the **Training Runs card** (both the run-detail header and the Presentation sub-tab) were plain ComboBoxes over the ~100-label Civitai catalog. This adds the same search-box UX the LoRA Viewer / Civitai Browser base-model filter flyouts already have — as a single-select picker.

## How

- **`SearchableBaseModelPicker`** (new UserControl): button face → flyout with search box + list. Case-insensitive substring narrowing (same rule as `RebuildFlyoutBaseModels`), search cleared on each open, picking an item closes the flyout and commits the selection.
- **`SearchableBaseModelPickerViewModel`** (new, plain + testable): per-instance engine. Tracks `INotifyCollectionChanged` sources live — the owning VMs `Clear()`+refill their label lists when the Civitai catalog resolves async — and unsubscribes from replaced sources.
- Four one-element XAML swaps; **zero view-model changes** — all three VMs already exposed `AvailableBaseModels` + `SelectedBaseModel`.
- The viewer/browser filter flyouts are **untouched** (they're multi-select filters; this is a picker — no shared code).

## Testing

- 9 new unit tests (TDD) on the engine: mirroring, narrowing, whitespace, selection+close, open-resets-search, live source tracking, ghost-unsubscribe, placeholder display.
- Full suite green: 5189 + 11 integration passed (one known pre-existing flake in `CivitaiBrowserFilterPersistenceTests` failed once, passes in isolation and on re-run).
- Manual GUI smoke still owed.

### DIFF
--- a/DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
+++ b/DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
@@ -1,0 +1,98 @@
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+
+namespace DiffusionNexus.UI.ViewModels;
+
+/// <summary>
+/// Per-instance engine behind the searchable base-model picker control. Single-select
+/// over a list of raw base-model labels, narrowed by a search box using the same rule
+/// as the viewer/browser filter flyouts (case-insensitive substring).
+/// </summary>
+public partial class SearchableBaseModelPickerViewModel : ObservableObject
+{
+    /// <summary>Narrows <see cref="VisibleItems"/> by case-insensitive substring.</summary>
+    [ObservableProperty]
+    private string _searchText = string.Empty;
+
+    /// <summary>The currently selected label, or <c>null</c> when nothing is picked.</summary>
+    [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(DisplayText))]
+    private string? _selectedItem;
+
+    /// <summary>Shown on the picker button while nothing is selected.</summary>
+    [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(DisplayText))]
+    private string _placeholderText = "Select base model…";
+
+    /// <summary>What the picker button shows: the selection, or the placeholder.</summary>
+    public string DisplayText => string.IsNullOrWhiteSpace(SelectedItem) ? PlaceholderText : SelectedItem;
+
+    /// <summary>Raised when a pick was made and the hosting flyout should close.</summary>
+    public event EventHandler? CloseRequested;
+
+    /// <summary>Called by the control when its flyout opens; a fresh open starts unfiltered.</summary>
+    public void OnFlyoutOpened() => SearchText = string.Empty;
+
+    /// <summary>Picks a label from the list and asks the flyout to close.</summary>
+    [RelayCommand]
+    private void Select(string label)
+    {
+        SelectedItem = label;
+        CloseRequested?.Invoke(this, EventArgs.Empty);
+    }
+
+    private IEnumerable<string>? _itemsSource;
+
+    /// <summary>
+    /// The full label list the picker offers. An <see cref="INotifyCollectionChanged"/>
+    /// source is tracked live — the real sources are <c>ObservableCollection&lt;string&gt;</c>
+    /// the owning view models Clear() and re-fill when the Civitai catalog resolves.
+    /// </summary>
+    public IEnumerable<string>? ItemsSource
+    {
+        get => _itemsSource;
+        set
+        {
+            if (ReferenceEquals(_itemsSource, value)) return;
+
+            if (_itemsSource is INotifyCollectionChanged oldObservable)
+            {
+                oldObservable.CollectionChanged -= OnItemsSourceCollectionChanged;
+            }
+
+            _itemsSource = value;
+
+            if (_itemsSource is INotifyCollectionChanged newObservable)
+            {
+                newObservable.CollectionChanged += OnItemsSourceCollectionChanged;
+            }
+
+            RebuildVisibleItems();
+        }
+    }
+
+    private void OnItemsSourceCollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
+        => RebuildVisibleItems();
+
+    /// <summary>The labels currently visible under the active search text.</summary>
+    public ObservableCollection<string> VisibleItems { get; } = [];
+
+    partial void OnSearchTextChanged(string value) => RebuildVisibleItems();
+
+    private void RebuildVisibleItems()
+    {
+        VisibleItems.Clear();
+        if (_itemsSource is null) return;
+
+        var search = SearchText.Trim();
+        foreach (var label in _itemsSource)
+        {
+            if (search.Length == 0 || label.Contains(search, StringComparison.OrdinalIgnoreCase))
+            {
+                VisibleItems.Add(label);
+            }
+        }
+    }
+}
--- a/DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
+++ b/DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
@@ -1,9 +1,27 @@
-using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
+using DiffusionNexus.UI.Utilities;
 
 namespace DiffusionNexus.UI.ViewModels;
+
+/// <summary>
+/// One row offered by the searchable base-model picker: the raw label plus whether
+/// it is the currently selected value (drives the selected-row highlight).
+/// </summary>
+public sealed partial class SearchableBaseModelPickerItem : ObservableObject
+{
+    public SearchableBaseModelPickerItem(string label, bool isSelected)
+    {
+        Label = label;
+        _isSelected = isSelected;
+    }
+
+    public string Label { get; }
+
+    [ObservableProperty]
+    private bool _isSelected;
+}
 
 /// <summary>
 /// Per-instance engine behind the searchable base-model picker control. Single-select
@@ -12,6 +30,10 @@ namespace DiffusionNexus.UI.ViewModels;
 /// </summary>
 public partial class SearchableBaseModelPickerViewModel : ObservableObject
 {
+    /// <summary>Default button caption while nothing is selected — single source for
+    /// both this VM's initial value and the control's styled-property default.</summary>
+    public const string DefaultPlaceholder = "Select base model…";
+
     /// <summary>Narrows <see cref="VisibleItems"/> by case-insensitive substring.</summary>
     [ObservableProperty]
     private string _searchText = string.Empty;
@@ -24,7 +46,14 @@ public partial class SearchableBaseModelPickerViewModel : ObservableObject
     /// <summary>Shown on the picker button while nothing is selected.</summary>
     [ObservableProperty]
     [NotifyPropertyChangedFor(nameof(DisplayText))]
-    private string _placeholderText = "Select base model…";
+    private string _placeholderText = DefaultPlaceholder;
+
+    /// <summary>
+    /// True only when an active (non-whitespace) search matches nothing. Stays false
+    /// while the source is merely empty (catalog still loading, or no source at all).
+    /// </summary>
+    [ObservableProperty]
+    private bool _hasNoMatches;
 
     /// <summary>What the picker button shows: the selection, or the placeholder.</summary>
     public string DisplayText => string.IsNullOrWhiteSpace(SelectedItem) ? PlaceholderText : SelectedItem;
@@ -34,6 +63,18 @@ public partial class SearchableBaseModelPickerViewModel : ObservableObject
 
     /// <summary>Called by the control when its flyout opens; a fresh open starts unfiltered.</summary>
     public void OnFlyoutOpened() => SearchText = string.Empty;
+
+    /// <summary>
+    /// Commits the single visible item, if the search has narrowed the list to exactly
+    /// one — Enter in the search box. Returns whether a pick was made.
+    /// </summary>
+    public bool TryCommitSingleMatch()
+    {
+        if (VisibleItems.Count != 1) return false;
+
+        Select(VisibleItems[0].Label);
+        return true;
+    }
 
     /// <summary>Picks a label from the list and asks the flyout to close.</summary>
     [RelayCommand]
@@ -47,8 +88,8 @@ public partial class SearchableBaseModelPickerViewModel : ObservableObject
 
     /// <summary>
     /// The full label list the picker offers. An <see cref="INotifyCollectionChanged"/>
-    /// source is tracked live — the real sources are <c>ObservableCollection&lt;string&gt;</c>
-    /// the owning view models Clear() and re-fill when the Civitai catalog resolves.
+    /// source is tracked live — the real sources are observable collections the owning
+    /// view models refill when the Civitai catalog resolves.
     /// </summary>
     public IEnumerable<string>? ItemsSource
     {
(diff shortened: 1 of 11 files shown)


DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (_itemsSource is INotifyCollectionChanged oldObservable)
+            {
+                oldObservable.CollectionChanged -= OnItemsSourceCollectionChanged;
+            }
+
+            _itemsSource = value;
+
+            if (_itemsSource is INotifyCollectionChanged newObservable)
+            {
+                newObservable.CollectionChanged += OnItemsSourceCollectionChanged;
```
Little-God1983: There's no teardown path for this subscription: the engine unsubscribes when the source is *replaced*, but nothing unhooks when a picker instance is discarded — no `OnDetachedFromLogicalTree`/`Unloaded` handling in the code-behind, and the subscription is strong (the replaced ComboBox tracked its ItemsSource weakly via the binding layer).

Today this is latent rather than live: all four hosts are either cached view instances (`ModuleItem.View` is reused on navigation, and the pickers are direct XAML content of those views) or `CreateTrainingRunDialog`, whose VM dies with the window. But the control is written to be reusable, and the first consumer that instantiates it inside a `DataTemplate` or a recreated view over a long-lived VM leaks one full picker subtree per recreation (collection → `OnItemsSourceCollectionChanged` → engine → ctor lambdas → control), with every dead engine still running a full `RebuildVisibleItems` on each catalog refresh.

Suggestion: in the control, override `OnDetachedFromLogicalTree` to set `Engine.ItemsSource = null` and re-apply from the styled property in `OnAttachedToLogicalTree` — or subscribe here via `WeakEventHandlerManager`.

DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            RebuildVisibleItems();
+        }
+    }
+
+    private void OnItemsSourceCollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
+        => RebuildVisibleItems();
```
Little-God1983: A full rebuild per `CollectionChanged` goes quadratic against how the sources are actually filled. All three owning VMs populate `AvailableBaseModels` with `Clear()` + ~80 individual `Add()`s (plus a possible `Insert`), and `ModelDetailViewModel.LoadAsync` re-runs that refill on **every tile selection** (`ModelDetailViewModel.cs:351` → `Editing.cs:341`). Each of those ~82 events triggers a full `VisibleItems.Clear()` + up-to-80 `Add()`s — roughly 3,300 collection notifications per tile click, each one processed by the live non-virtualized `ItemsControl` once the flyout has been realized (and doubled on the training-run card, where the header picker and the Presentation sub-tab picker both watch the same collection).

Two cheap fixes that compose:
- make `VisibleItems` a `BatchObservableCollection<string>` and rebuild via `ReplaceAll(...)` — the helper in `Utilities/BatchObservableCollection.cs` that `LoraViewerViewModel.FlyoutBaseModels` already uses for exactly this shape. One `Reset` per rebuild instead of Reset + N Adds; also improves the per-keystroke path.
- collapse the refill burst: switch the source collections to `ReplaceAll` too (one event per catalog refresh for every subscriber), which keeps this VM free of dispatcher/debounce machinery.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
```diff
@@ -0,0 +1,85 @@
...
+  <Button x:Name="PickerButton"
+          x:DataType="vm:SearchableBaseModelPickerViewModel"
+          HorizontalAlignment="Stretch"
+          HorizontalContentAlignment="Stretch"
+          Padding="8,6"
+          Background="#2A2A2A"
+          BorderBrush="#555"
+          BorderThickness="1"
+          CornerRadius="4"
+          ToolTip.Tip="{Binding DisplayText}">
```
Little-God1983: This hardcoded `ToolTip.Tip` shadows any tooltip a call site sets on the control: the button covers the entire UserControl, and Avalonia resolves the tip from the innermost element that has one. Concretely, `ModelDetailView.axaml` still passes `ToolTip.Tip="Override the base model for the currently selected version (saved per-version)"` — the one call site where a pick immediately persists a per-version override — and that explanation is now unreachable; hovering shows only the selected label echoed back. The neighboring source-badge tooltips even direct users to this dropdown for that explanation.

Suggestion: surface a tip the consumer can set (or only apply the DisplayText tip when the text is actually trimmed, as an overflow aid).

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
```diff
@@ -0,0 +1,85 @@
...
+                 VerticalAlignment="Center"/>
+    </Grid>
+    <Button.Flyout>
+      <Flyout Placement="BottomEdgeAlignedLeft">
+        <Border CornerRadius="4"
+                Padding="8"
+                MinWidth="240"
+                MaxHeight="420">
+          <Grid RowDefinitions="Auto,*">
+            <TextBox x:Name="SearchBox"
```
Little-God1983: Keyboard and screen-reader paths regress vs the replaced ComboBox, and nothing here re-establishes them:

- Enter in the search box commits nothing — type "turbo" until exactly one item is visible, press Enter, and the flyout just stays open;
- arrow keys can't move focus from the TextBox into the item buttons (a plain StackPanel of Buttons has no directional navigation), so a keyboard-only pick means tabbing button-by-button;
- the closed control lost ComboBox's type-ahead / arrow-key value cycling;
- the ComboBox automation peer (role, current value, expand state) becomes an unnamed Button whose content is a Grid — no `AutomationProperties` anywhere, so a screen reader announces an empty "button" with no current value.

Minimum worth doing: commit the single visible match on Enter, and set `AutomationProperties.Name` (bound to `DisplayText`) on the button.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
```diff
@@ -0,0 +1,85 @@
...
+                     Text="{Binding SearchText, Mode=TwoWay}"
+                     Watermark="Search base models..."
+                     Margin="4,0,4,8"/>
+
+            <ScrollViewer Grid.Row="1"
+                          MaxHeight="340"
+                          HorizontalScrollBarVisibility="Disabled"
+                          VerticalScrollBarVisibility="Auto">
+              <StackPanel>
+                <TextBlock Text="No matches."
```
Little-God1983: "No matches." binds only `!VisibleItems.Count`, so it also shows when the source is simply empty — `ItemsSource` still null, or the async Civitai catalog fetch in flight or failed. `CreateTrainingRunDialog` fires its fetch on open, so clicking the picker straight away shows "No matches." with an empty search box: a loading state misreported as a failed search. Worth distinguishing — e.g. only show this when `SearchText` is non-empty, or add a loading/empty-source state.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
```diff
@@ -0,0 +1,85 @@
...
+      <TextBlock Grid.Column="1"
+                 Text="&#9662;"
+                 FontSize="10"
+                 Opacity="0.6"
+                 Margin="8,0,0,0"
+                 VerticalAlignment="Center"/>
+    </Grid>
+    <Button.Flyout>
+      <Flyout Placement="BottomEdgeAlignedLeft">
+        <Border CornerRadius="4"
```
Little-God1983: Three chrome details diverge from the two flyouts this control deliberately mimics (`LoraViewerView.axaml:59`, `CivitaiBrowserView.axaml`):

- their Border sets `Background="#1E1E1E" BorderBrush="#444" BorderThickness="1"`; this one sets none of them, so the popup renders with theme-default presenter chrome and looks different from the filter flyouts it replicates;
- the popup is content-sized (`MinWidth="240"`) and doesn't track the button width — at stretched call sites like the detail card the dropdown opens visibly narrower than the button, where the old ComboBox dropdown matched the control's width;
- nit: `MaxHeight="420"` here can never bind — content maxes out around TextBox + ScrollViewer(340) + padding ≈ 396, so the ScrollViewer's cap is the only one that governs. One cap would do.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
```diff
@@ -0,0 +1,85 @@
...
+                          MaxHeight="340"
+                          HorizontalScrollBarVisibility="Disabled"
+                          VerticalScrollBarVisibility="Auto">
+              <StackPanel>
+                <TextBlock Text="No matches."
+                           FontSize="12"
+                           Opacity="0.6"
+                           Margin="4,2"
+                           IsVisible="{Binding !VisibleItems.Count}"/>
+                <ItemsControl ItemsSource="{Binding VisibleItems}">
```
Little-God1983: The list gives no indication of the current selection and never scrolls it into view — the ComboBox opened with the active item highlighted and in view. With ~80 identically-styled rows and near-duplicate labels ("Wan Video 2.2 T2V-A14B" vs "Wan Video 2.2 TI2V-5B"), a user opening the picker to check or adjust a value can't see which entry is active without closing the flyout and reading the button — and on the detail card a mis-pick is immediately saved per-version. A selected-state style on the item button (compared against the engine's `SelectedItem`) plus a scroll-into-view on open would restore this.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml.cs
```diff
@@ -0,0 +1,113 @@
...
+        AvaloniaProperty.Register<SearchableBaseModelPicker, string?>(
+            nameof(SelectedItem),
+            defaultBindingMode: BindingMode.TwoWay);
+
+    /// <summary>
+    /// Defines the <see cref="PlaceholderText"/> property.
+    /// </summary>
+    public static readonly StyledProperty<string> PlaceholderTextProperty =
+        AvaloniaProperty.Register<SearchableBaseModelPicker, string>(
+            nameof(PlaceholderText), "Select base model…");
```
Little-God1983: This default is defined twice — here and as the engine's field initializer (`SearchableBaseModelPickerViewModel.cs:27`) — and the copy here is effectively dead: `OnPropertyChanged` only fires on *changes*, so the registration default never flows into the engine, and a call site that sets nothing (ModelDetailView) renders the VM's copy only because the two strings happen to match today. Edit one of them and `GetValue(PlaceholderTextProperty)` and the rendered UI silently disagree.

Single-source it: a `public const string DefaultPlaceholder` on the VM referenced here, or seed `Engine.PlaceholderText = PlaceholderText;` once in the ctor.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// <summary>
+    /// Per-instance engine holding the search/narrowing/selection state. Exposed for the
+    /// control's own XAML; consumers bind the styled properties instead.
+    /// </summary>
+    public SearchableBaseModelPickerViewModel Engine { get; } = new();
```
Little-God1983: The doc comment doesn't match the code: the control's XAML never references `Engine` (the DataContext is assigned in the ctor), and nothing else touches it. As public API it invites a consumer to write `picker.Engine.SelectedItem = ...`, bypassing the styled-property sync and creating a second write path. A `private readonly` field (with the comment fixed) would close that door.

DiffusionNexus.UI/Views/Tabs/TrainingRunsTabView.axaml
```diff
@@ -85,13 +86,12 @@
...
           <TextBlock Grid.Column="1" Text="{Binding SelectedTrainingRun.Name}" FontSize="16" FontWeight="SemiBold" VerticalAlignment="Center"/>
-          <ComboBox Grid.Column="2"
+          <controls:SearchableBaseModelPicker Grid.Column="2"
                     ItemsSource="{Binding SelectedTrainingRun.AvailableBaseModels}"
                     SelectedItem="{Binding SelectedTrainingRun.SelectedBaseModel, Mode=TwoWay}"
                     PlaceholderText="Base model…"
-                    Margin="12,0,0,0" Padding="8,5" FontSize="13"
-                    VerticalAlignment="Center" Background="#2A2A2A"
-                    BorderBrush="#555" CornerRadius="4" MinWidth="160"/>
+                    Margin="12,0,0,0" FontSize="13"
```
Little-God1983: Minor: the dropped `Padding="8,5"` can't be re-established from the call site — UserControl padding doesn't reach the inner button, which hardcodes `8,6` — so the picker sits 2px taller than the trigger-words TextBox and the buttons in this header row. If the row alignment matters, the control needs to forward padding (or the inner button could use `8,5`).

Little-God1983 (commented):
Reviewed at a58d6366. The engine design is a real improvement over the ComboBox in one important way: selection is decoupled from `ItemsSource`, so the old "Clear() nulls the selection through the TwoWay binding" failure the `_suppressBaseModelSave` guards were fighting can't happen anymore. The 9 engine tests are solid and match the implementation.

10 inline comments below. The two most substantial: the missing unsubscribe path (latent leak for a control written to be reused) and the shadowed call-site tooltip on the detail card. Then: O(n²) rebuilds on catalog refill, keyboard/screen-reader regressions vs ComboBox, no selected-item indication in the list, "No matches." during async load, flyout chrome/width drift, a duplicated placeholder default, and two smaller nits.

Out-of-diff notes (nothing to anchor these to):
- `TrainingRunCardViewModel.cs:540-542` — the `OnPropertyChanged(nameof(SelectedBaseModel))` re-raise "so the ComboBox re-binds" is now dead code: no ComboBox binds `SelectedBaseModel` anymore, and the picker never de-selects on list refresh. Worth deleting (with its comment) next time that file is touched.
- The narrowing rule ("same rule as `RebuildFlyoutBaseModels`") is now hand-copied a third time, coupled only by prose. A small shared predicate helper would keep the three base-model search boxes answering identically if the rule ever changes.
- Trivial: `xmlns:local` in SearchableBaseModelPicker.axaml is unused, and PresentationSubTabView passes `PlaceholderText="Select base model…"`, which restates the built-in default verbatim.

DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
Little-God1983: Fixed in d81650a8. OnDetachedFromLogicalTree nulls the engine ItemsSource (dropping the CollectionChanged subscription), OnAttachedToLogicalTree re-applies it from the styled property — so cached views also stop rebuilding while off-screen.

DiffusionNexus.UI/ViewModels/SearchableBaseModelPickerViewModel.cs
Little-God1983: Fixed in d81650a8, both halves: VisibleItems is now a BatchObservableCollection rebuilt via ReplaceAll (one Reset per change), and all three owning VMs refill AvailableBaseModels via ReplaceAll too — the persisted-value Insert is folded into the rebuilt list, final order unchanged (existing order-asserting tests still pass). The dead SelectedBaseModel re-raise in TrainingRunCardViewModel went with it.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
Little-God1983: Fixed in d81650a8 — the internal ToolTip.Tip is removed, so the call-site tip on ModelDetailView resolves again. No trimmed-only echo added; nothing needs it today.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
Little-God1983: The minimum is in d81650a8: Enter in the search box commits the single visible match (TryCommitSingleMatch on the engine, covered by tests), and the button carries AutomationProperties.Name bound to DisplayText. Arrow-key navigation into the list and closed-state value cycling are not restored.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
Little-God1983: Fixed in d81650a8 — the engine exposes HasNoMatches, true only when a non-whitespace search yields nothing; an empty or still-loading source no longer shows "No matches." (both states covered by tests).

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
Little-God1983: Fixed in d81650a8 — the flyout Border now sets #1E1E1E / #444 / 1 like the filter flyouts, its MinWidth is raised to the button width on every open, and the dead MaxHeight=420 is gone (the ScrollViewer 340 cap governs alone).

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml
Little-God1983: Fixed in d81650a8 — rows are now small items (Label + IsSelected) so the active row gets a selected style (#335599 fill, #5577BB border), and the control scrolls it into view after the flyout opens. Selection-flag behavior is engine-tested.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml.cs
Little-God1983: Fixed in d81650a8 — DefaultPlaceholder const on the engine is the single source, referenced by both the field initializer and the styled-property registration.

DiffusionNexus.UI/Views/Controls/SearchableBaseModelPicker.axaml.cs
Little-God1983: Fixed in d81650a8 — Engine is now a private readonly field and the stale comment is corrected.

DiffusionNexus.UI/Views/Tabs/TrainingRunsTabView.axaml
Little-God1983: Fixed in d81650a8 by setting the inner button to 8,5 (all four sites now share the TextBox height). Forwarding the control Padding was the wrong tool here: UserControl default template already applies Padding around its content, so a consumer-set value would double-apply as an outer inset plus inner padding.

PR #554, opened 2026-09-01
Remove the dead write-only BaseModelType enum (column kept, no migration) + UI reusables catalog

### LINKED ISSUE #553: Remove the dead write-only BaseModelType enum, column and index
Spun out of #526 (closed as obsolete — see the closing comment there for the full verification).

`BaseModelType` is write-only dead code: every non-test reference writes it (or is a parse helper feeding a write); nothing reads it. All real base-model filtering/selection uses the raw string labels from `ICivitaiBaseModelCatalog`. Because the enum's Civitai mapping is ~65 labels behind and will keep drifting forever, the honest fix is removal, not maintenance.

Scope:
- Delete the `BaseModelType` enum and `BaseModelTypeExtensions.ParseCivitai` (plus their tests).
- Drop the `ModelVersion` column that stores it and its index (`ModelVersionConfiguration`), with a migration.
- Remove the write sites that stamp it.
- The 102 stale rows disappear with the column.

Low risk — the column is TEXT and nothing reads it — but the migration touches the shipped `Diffusion_Nexus-core.db` schema, so it should ride an ordinary schema-bump release, not a hotfix.

Little-God1983:
Closes #553.

## The cleanup (#553)

`BaseModelType` was a **hardcoded list of 24 base-model names that nothing ever read**. Every non-test reference wrote it, or was a parse helper feeding a write — all real base-model filtering and selection goes through the raw string labels from `ICivitaiBaseModelCatalog`. Its Civitai mapping was ~65 labels behind and would keep drifting forever, so removal is the honest fix.

I re-verified write-only status rather than trusting the issue: the only three ambiguous `.BaseModel` reads in prod code turned out to be Civitai DTO strings (`CivitaiModelVersion.BaseModel`, `CivitaiVersionTabItem.BaseModel`), not the entity enum. The single entity read is `LoraDownloadService` copying it from one version to another — itself part of a write path.

Removed:
- the `BaseModelType` enum and `BaseModelTypeExtensions.ParseCivitai` (+ tests)
- the eight write sites that stamped it, and the two `ParseBaseModel` wrappers
- the `SidecarMetadataApplier.WriteBaseModel` wrapper — the four sites call `BaseModelWriter.Write` directly, so the blank-is-not-an-answer rule has one home

`BaseModelWriter` survives — it still holds the rule that a blank upstream answer is a missing answer, not an instruction to forget the stored one. Its docs no longer claim to keep two spellings in sync, because there is only one now. The hash-fallback backfill in `LoraDownloadService` now goes through it too (it was the last write that could blank a stored base model).

## The column stays — no migration (scope change after review)

The first version of this PR dropped `ModelVersions.BaseModel` and its index. Review showed that would have been **the first destructive migration in the project's history**, and the recovery service is built on the assumption that migrations only add things: it stamps failed migrations as applied without running them, un-stamps history rows the running binary doesn't know about, and its repairs only add columns. Dropping the column made the schema forward-only — a user rolling back to an earlier installer would hit `no such column: BaseModel` on the first LoRA query, and the code that fails lives in binaries already shipped, so nothing in this PR could defuse it.

So the column is kept as dead weight and there is **no migration**: snapshot, `schema.sql` and `DatabaseRecoveryService` are untouched vs `develop`.

- `ModelVersion.BaseModel` is now a plain `string` that always holds `"Unknown"`, marked `[Obsolete(error: true)]` so nothing can compile against it. `"Unknown"` was a member of the old enum, so every earlier build's enum converter still parses it.
- `ModelVersionConfiguration` maps it **by name** (`Property<string>("BaseModel")`, `HasIndex("BaseModel")`) with the exact snapshot shape — TEXT NOT NULL, max 50, indexed.
- The 102 stale rows keep their old enum names; nothing reads them and every one of them is parseable by old builds.

### Test notes

Every dropped enum assertion sat directly beside an equivalent `BaseModelRaw` assertion, so no behavioural coverage was lost.

`SyncSchemaMigrationTests`' additive-only guard is back in its strict form (it correctly fired on the drop in the first version). Two new tests pin the decision:

- `ModelHasNoPendingChangesAgainstSnapshot` — the production DbContext registration ignores `PendingModelChangesWarning`, so a model that drifts from the snapshot would migrate silently and surface as a runtime SQL error for users. This is the guard that warning would have been.
- `EfInsertStillFillsTheKeptBaseModelColumnWithUnknown` — an EF-authored `ModelVersion` insert lands `"Unknown"` in the kept NOT NULL column without any production code touching the property.

`Persist_DuplicateVersionBlankUpstreamBaseModel_KeepsStoredOne` reproduces the hash-fallback blanking (`""` over `"Pony"`) that the `BaseModelWriter` change fixes.

## The catalog (folded in, as a cleanup companion)

`copilot-instructions.md` already required looking for reusable components before building new UI, but there was nothing to look at. `DiffusionNexus.UI/REUSABLES.md` now catalogs the reuse surface: base classes, the ~25 controls in `Views/Controls/`, the `IDialogService` dialog surface, converters, helpers. The rule now names the file and asks for a row when a new reusable piece is added.

It also records the one real gap: there is no shared style dictionary, so all 95 axaml files declare brushes and paddings inline and drift.

## Verification

- `dotnet build DiffusionNexus.sln` — 0 errors (12 warnings, all pre-existing in untouched files)
- `dotnet test DiffusionNexus.sln` — **5178 + 11 passed, 0 failed**, 2 skipped (online canaries)
- No schema change: `DiffusionNexus.DataAccess/Migrations`, `schema.sql` and `Recovery/` have no diff vs `develop`


DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
```diff
@@ -0,0 +1,39 @@
...
+    public partial class RemoveDeadBaseModelTypeColumn : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropIndex(
+                name: "IX_ModelVersions_BaseModel",
+                table: "ModelVersions");
+
+            migrationBuilder.DropColumn(
```
Little-God1983: **The project's first destructive migration meets a recovery service that can only add columns.**

`DatabaseRecoveryService.InitializeAndRepair` catches *every* `Migrate()` failure ("already exists", "duplicate column", "no such column", and a generic `catch (Exception)`) and responds with `CheckAndRepairSchema` + `MarkPendingMigrationsAsApplied` — which stamps `__EFMigrationsHistory` **without running the migration body**.

Concrete failure: a user DB where `Migrate()` throws for any reason (locked file, an earlier partial migration, a pre-existing schema gap). `RemoveDeadBaseModelTypeColumn` gets stamped as applied, but `ModelVersions.BaseModel` is still physically there as `TEXT NOT NULL` **with no default** (`20251222195309_InitialCreate.cs:111`). EF's model no longer knows the property, so every subsequent `INSERT INTO ModelVersions` omits it:

```
SQLite Error 19: 'NOT NULL constraint failed: ModelVersions.BaseModel'
```

Every model discovery, LoRA download and sync that creates a version fails from then on, permanently. `CheckAndRepairSchema` cannot heal it — `RepairModelsTableColumns` / `RepairModelImagesTableColumns` / `EnsureModelSyncStatesTable` only ever *add* things; nothing drops a leftover column.

Until this PR the stamp-without-run shortcut was safe precisely because every migration was additive. Worth either teaching `CheckAndRepairSchema` to drop a leftover `ModelVersions.BaseModel`, or making `MarkPendingMigrationsAsApplied` refuse to stamp migrations that are not add-only.

DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace DiffusionNexus.DataAccess.Migrations.Core
+{
+    /// <inheritdoc />
+    public partial class RemoveDeadBaseModelTypeColumn : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropIndex(
```
Little-God1983: **Rolling back to an earlier build now bricks the database, where it used to be harmless.**

`DatabaseRecoveryService.CleanStaleMigrationHistory` deletes any `__EFMigrationsHistory` row whose migration id is not in `dbContext.Database.GetMigrations()` — i.e. it silently un-stamps migrations the running assembly does not know about.

Sequence: user updates → this migration runs, `ModelVersions.BaseModel` is physically gone → user rolls back to the previous release (a normal thing for a desktop app) → that build's `CleanStaleMigrationHistory` removes the `20260901222123_RemoveDeadBaseModelTypeColumn` row → its model still maps `ModelVersion.BaseModel` → every ModelVersions query fails with `no such column: BaseModel`. The `no such column` handler runs `CheckAndRepairSchema` (there is no `RepairModelVersionsTableColumns`) and `MarkPendingMigrationsAsApplied` (nothing pending), so nothing recovers it.

Every migration before this one was additive, so a downgrade left an unused extra column and the app kept working. This is the first one where forward-only is a hard requirement, and neither the recovery service nor the release notes say so. At minimum this deserves a `RepairModelVersionsTableColumns` entry that re-adds `BaseModel TEXT NOT NULL DEFAULT 'Unknown'`, mirroring the two repair tables that already exist for exactly this class of problem.

DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "BaseModel",
+                table: "ModelVersions",
+                type: "TEXT",
+                maxLength: 50,
+                nullable: false,
+                defaultValue: "");
```
Little-God1983: **`Down()` is not actually reversible — `defaultValue: ""` is not a valid `BaseModelType` name.**

The column was mapped with `entity.Property(e => e.BaseModel).HasConversion<string>()`, so its values were enum *names* (`"Unknown"`, `"Pony"`, …). Re-adding it filled with `""` means every existing row gets a value that the pre-drop code cannot parse: the moment an older build materializes any `ModelVersion`, EF's enum-to-string converter calls `Enum.Parse` on `""` and throws — so the rollback this method exists to enable fails on the first LoRA query.

```suggestion
            migrationBuilder.AddColumn<string>(
                name: "BaseModel",
                table: "ModelVersions",
                type: "TEXT",
                maxLength: 50,
                nullable: false,
                defaultValue: "Unknown");
```

DiffusionNexus.UI/Services/LoraDownloadService.cs
```diff
@@ -635,7 +634,6 @@ public async Task<MetadataPersistOutcome> PersistDownloadedModelAsync(string fil
                     {
                         duplicateVersion.Name = version.Name;
                         duplicateVersion.Description ??= version.Description;
-                        duplicateVersion.BaseModel = version.BaseModel;
                         duplicateVersion.BaseModelRaw = version.BaseModelRaw;
```
Little-God1983: **This is now the only base-model write site that bypasses `BaseModelWriter`, and it can blank a known base model.**

The PR removed the `duplicateVersion.BaseModel = version.BaseModel;` line right above, which leaves this unconditional assignment as the sole survivor of the rule the rest of the PR re-affirms: *"a blank upstream answer is a missing answer, not an instruction to forget the stored one"* (`BaseModelWriter.Write`, `SidecarMetadataApplier.WriteBaseModel`, `CivitaiMetadataApplier:170`).

`CivitaiModelVersion.BaseModel` is `public string BaseModel { get; init; } = string.Empty;`, and line 510 copies it verbatim into `version.BaseModelRaw`. So: a duplicate matched by hash, `duplicateVersion.IsUserEdited == false`, `duplicateVersion.BaseModelRaw == "Pony"`, and a Civitai response that omits `baseModel` → this line writes `""` and the version drops into the sorter's Unknown bucket, undoing whatever the sidecar/header/heuristic rungs had established. `Description`, `DownloadUrl` and `PublishedAt` on the neighbouring lines all use `??=` precisely to avoid this shape.

```suggestion
                        BaseModelWriter.Write(duplicateVersion, version.BaseModelRaw);
```
(or `if (!string.IsNullOrWhiteSpace(version.BaseModelRaw))` if you would rather not take the Service dependency here)

DiffusionNexus.Tests/Sync/DataAccess/SyncSchemaMigrationTests.cs
```diff
@@ -73,7 +73,19 @@ public void MigrationIsAdditiveAndUppercasesHashes()
...
+            }
+
+            // ModelVersions is the one table the additive-only rule no longer covers verbatim.
+            // #553 dropped its write-only BaseModel column on purpose, and dropping a column on
+            // SQLite rebuilds the table, so the surviving columns come back re-ordered rather than
+            // in their original positions. The invariant that still holds — and the one that
+            // actually matters — is that nothing BUT that column went missing.
+            after["ModelVersions"].Should().BeEquivalentTo(
+                before["ModelVersions"].Where(c => c != "BaseModel:TEXT:1"),
+                "the only column ModelVersions may lose is the one #553 dropped deliberately");
```
Little-God1983: **The replacement assertion is tighter than the guard it replaces — it now forbids *additive* changes to `ModelVersions` too.**

`StartWith(cols)` allowed appended columns (that was the point: "additive-only"). `BeEquivalentTo` is exact set equality, so the next migration that legitimately **adds** a column to `ModelVersions` will fail `MigrationIsAdditiveAndUppercasesHashes` with "expected collection to be equivalent", and the fix will not be obvious to whoever hits it — the message and the comment both talk about a *dropped* column.

Keeping the order-insensitivity you need while restoring additivity:

```suggestion
            after["ModelVersions"].Should().Contain(
                before["ModelVersions"].Where(c => c != "BaseModel:TEXT:1"),
                "the only column ModelVersions may lose is the one #553 dropped deliberately");
            after["ModelVersions"].Should().NotContain("BaseModel:TEXT:1");
```

DiffusionNexus.Service/Services/Sync/SidecarMetadataApplier.cs
```diff
@@ -501,17 +501,15 @@ private static bool ApplySimpleJsonFormat(
     private static bool CanWriteVersionText(ModelVersion dbVersion) => !dbVersion.IsUserEdited;
 
     /// <summary>
-    /// Writes both spellings of the base model — the raw upstream string and the
-    /// <see cref="BaseModelType"/> the viewer's filter reads — using the same parser the detail
-    /// view's editor uses. Writing only the raw one left the enum reporting the previous base
-    /// model forever. Callers must check <see cref="CanWriteVersionText"/> first.
+    /// Writes the version's base model — the raw upstream string, the only spelling the app
+    /// stores. Callers must check <see cref="CanWriteVersionText"/> first.
```
Little-God1983: **Four documentation files still describe the deleted enum as live.** The PR carefully updated the XML docs and `schema.sql`, but the prose docs were missed:

- `DiffusionNexus.UI/Doc/LoraViewer.md:307-310` — *"Every write of `BaseModelRaw` also writes the `BaseModel` enum (`BaseModelTypeExtensions.ParseCivitai`, the same helper the editor uses), so the viewer's base-model filter and the label the detail view shows never disagree"*. This is now false in both halves, and it is the doc a future reader consults before touching this exact code path.
- `DiffusionNexus.UI/Doc/LoraViewer.md:655` — lists `BaseModelTypeExtensions` as a live helper in the components table.
- `DiffusionNexus.Domain/README.md:103,115` and its verbatim duplicate `docs/V2/Domain.md:103,115` — `ModelVersion` sample code (`BaseModel = BaseModelType.SDXL10`, `BaseModel = BaseModelType.WanVideo22`) that no longer compiles, in the README of the very assembly that lost the type; `:160` lists `BaseModelType` in the Enums table.
- `Doc/TestCoverageEvaluation.md:36,135,330` — cites `BaseModelTypeExtensionsTests`, deleted by this PR.

This matters more than usual here: the same PR adds a rule to `copilot-instructions.md` about keeping a catalog current in the same commit.

DiffusionNexus.Service/Services/Sync/SidecarMetadataApplier.cs
```diff
@@ -501,17 +501,15 @@ private static bool ApplySimpleJsonFormat(
...
     /// </summary>
     /// <returns>Whether anything was written.</returns>
     /// <remarks>
     /// A blank <c>baseModel</c> is a missing answer, not an instruction to forget the stored one
     /// (F6). The call sites only rejected an <i>absent</i> key, so a sidecar carrying
-    /// <c>"baseModel": ""</c> blanked the raw string and set the enum to <c>Unknown</c> on a version
-    /// nobody had edited. The check lives here so all four of them inherit it. Delegates to
+    /// <c>"baseModel": ""</c> blanked the stored base model on a version nobody had edited. The
+    /// check lives here so all four of them inherit it. Delegates to
```
Little-God1983: **Dead indirection plus a comment that now contradicts itself.**

With the enum gone, `WriteBaseModel` is a pure one-line pass-through to `BaseModelWriter.Write` carrying a 13-line doc block. The block still claims *"The check lives here so all four of them inherit it"* — but the check does not live here any more; it lives in `BaseModelWriter.Write`, which the next sentence correctly says it delegates to. A reader auditing the blank-is-not-an-answer rule will look in the wrong place.

Either call `BaseModelWriter.Write` directly from the four sites (`:275`, `:398`, `:464`, `:475`) and delete this wrapper, or keep the wrapper and reword the remark so the rule is attributed to `BaseModelWriter`.

.github/copilot-instructions.md
```diff
@@ -11,7 +11,7 @@ In this project use the Nuget packages DiffusionNexus.Installer.SDK from its ori
 - Before any Database Entity classes, IEntityTypeConfigurations or the database migrations are modified execute the publish.ps1 script to make sure there is a last backup of the app with a working database.
 - When A Keyboard Shortcut is Added make sure its documented in a file called DiffusionNexus.UI\\Doc\\Shortcuts.md
 - When fixing a failing Unit test, check thoroughly if its the unit test that is in need of fixing or if its the code that has an actual bug.
-- Every time new UI elements are added or UI is built from the ground up, always look first for reusable components in the project before creating new ones.
+- Every time new UI elements are added or UI is built from the ground up, always look first for reusable components in the project before creating new ones. The catalog of reusable controls, base classes, dialogs, converters and helpers is `DiffusionNexus.UI\REUSABLES.md` — check it before writing a new control, and add a row to it in the same commit whenever you create a new reusable piece.
```
Little-God1983: **Two rules from this very file appear to be broken by this PR.**

1. `.github/copilot-instructions.md:10` — *"When a refactoring task is issued then mark the method/function/class you are refactoring as obsolete and to be removed later to avoid code duplication/dead code."* Commit `046fde2` is titled `refactor: remove the dead write-only BaseModelType enum, column and index` and deletes the public `BaseModelType`, `BaseModelTypeExtensions.ParseCivitai` and `ModelVersion.BaseModel` from the shared `DiffusionNexus.Domain` assembly in one step, with no `[Obsolete]` window. Given the migration is what makes the removal irreversible, a release that ships `[Obsolete]` first and drops the column second would also defuse the downgrade hazard flagged on the migration.

2. `.github/copilot-instructions.md:11` — *"Before any Database Entity classes, IEntityTypeConfigurations or the database migrations are modified execute the publish.ps1 script to make sure there is a last backup of the app with a working database."* This PR modifies all three (`ModelVersion.cs`, `ModelVersionConfiguration.cs`, a new migration) and is the first migration in the project's history that destroys data. Nothing in the PR records that `publish.ps1` ran — worth confirming before merge, since this is exactly the case the rule was written for.

DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
Little-God1983: Fixed in e9045c18. Confirmed against the code: all five catch paths stamp via MarkPendingMigrationsAsApplied without running bodies, and every existing repair is add-only. Added DropLeftoverModelVersionsBaseModelColumn to CheckAndRepairSchema — when ModelVersions.BaseModel is still physically present it drops IX_ModelVersions_BaseModel first (SQLite refuses to drop an indexed column), then the column. Two new tests reproduce the stamped-without-run state: one proves the exact EF INSERT shape (column omitted) works after repair, one proves rows survive the drop.

DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
Little-God1983: Fixed in e9045c18 — Down() now uses defaultValue "Unknown". Verified the failure mode: the pre-drop mapping is HasConversion<string>(), so "" would throw in the enum converter on the first materialized ModelVersion.

DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
Little-God1983: The mechanics check out — old builds' CleanStaleMigrationHistory does un-stamp this row, and their repair set has nothing for ModelVersions. But the suggested RepairModelVersionsTableColumns can't defuse it: the failing code path runs in the OLD binary, which is already shipped and frozen — a repair added in THIS build never executes during a downgrade. What e9045c18 does instead: (1) Down() is now genuinely reversible (default 'Unknown'), so a deliberate rollback has a working path; (2) Up() uses DROP INDEX IF EXISTS, so the downgrade→re-upgrade round trip (old build un-stamps, new build replays the migration against a schema where the drop already happened) recovers instead of throwing on DropIndex; (3) the migration header documents the forward-only requirement and the manual re-add statement. The only full defusal would be a two-release staging (release N unmaps the property but keeps the column, N+1 drops it) — flagged to Chris as a release-planning decision, since #553 scoped a single release.

DiffusionNexus.UI/Services/LoraDownloadService.cs
Little-God1983: Fixed in e9045c18 — the write goes through BaseModelWriter.Write (the file already imports DiffusionNexus.Service.Services.Sync, so no new dependency). One precision on scope: an id-matched duplicate never reaches this write — the linkage+text block is gated on duplicateVersion.CivitaiId is null, so linked versions are frozen. The exposed path is exactly the hash-fallback orphan backfill your comment describes: local-discovery version, base model established by a sidecar/header rung, Civitai response omitting baseModel. Regression test Persist_DuplicateVersionBlankUpstreamBaseModel_KeepsStoredOne reproduced the blanking ("" over "Pony") before the fix and pins it now.

DiffusionNexus.Tests/Sync/DataAccess/SyncSchemaMigrationTests.cs
Little-God1983: Fixed in e9045c18 as suggested — Contain (superset) restores additivity for future ModelVersions columns, plus an explicit NotContain("BaseModel:TEXT:1") so the drop itself stays pinned. Comment now says why set-not-prefix (SQLite table rebuild re-orders surviving columns).

DiffusionNexus.Service/Services/Sync/SidecarMetadataApplier.cs
Little-God1983: Fixed in e9045c18: LoraViewer.md §sync (the enum-write claim — now states BaseModelRaw is the only stored spelling) and its components table; Domain/README.md and docs/V2/Domain.md samples (now BaseModelRaw = "SDXL 1.0" / "Wan Video 2.2") and the enums table; TestCoverageEvaluation.md's three mentions. Left untouched: the .superpowers/sdd briefs/reports and the dated 2026-08-23 plan — those are point-in-time records of past rulings, and rewriting them would falsify history rather than fix documentation.

DiffusionNexus.Service/Services/Sync/SidecarMetadataApplier.cs
Little-God1983: Deleted the wrapper in e9045c18 — the four sites call BaseModelWriter.Write directly, so the blank-is-not-an-answer rule has exactly one home and the audit trail points there. The F6 sidecar history survives in BaseModelWriter.Write's remarks and the SidecarMetadataApplierTests comments.

.github/copilot-instructions.md
Little-God1983: On rule 2 (publish.ps1): Chris waived it for this PR in-session — a fresh backup was taken immediately before the work started. Noting it here so the record exists.

On rule 1 (obsolete-first): partially disputed. The rule's stated rationale is avoiding duplication between an old API and its replacement during a refactoring window — here there is no replacement; #553 (authored by Chris) explicitly scoped outright deletion of dead code. An [Obsolete] pass on the enum alone would also NOT defuse the downgrade hazard: that hazard comes from the old build's EF model mapping the column, and it persists whether or not the members carry [Obsolete]. The variant that WOULD defuse it is a two-release staging — release N keeps the column but removes the property from the EF model, release N+1 drops the column — at the cost of shipping the dead column one more release. That trade-off is Chris's call and is flagged in the PR summary; the migration header documents the forward-only requirement either way.

DiffusionNexus.DataAccess/Migrations/Core/20260901222123_RemoveDeadBaseModelTypeColumn.cs
Little-God1983: Resolved by scope change in 4f8b1511: the column drop is withdrawn. ModelVersions.BaseModel stays as a dead column — a plain string property that always holds "Unknown" (a member of the old enum, so every earlier build's converter parses it), marked [Obsolete(error: true)] and mapped by name so nothing compiles against it. There is no migration: snapshot, schema.sql and DatabaseRecoveryService are identical to develop again, and the additive-only guard is back in its strict form. New ModelHasNoPendingChangesAgainstSnapshot pins the model against the snapshot (the production DbContext ignores PendingModelChangesWarning, so nothing else would catch drift); EfInsertStillFillsTheKeptBaseModelColumnWithUnknown proves an EF insert still fills the NOT NULL column. The enum deletion and the eight write-site removals are unchanged.

PR #555, opened 2026-09-03
feat(editor): Canvas Extend tool

Little-God1983:
## Summary

A **Canvas Extend** tool for the Image Editor: grow the canvas around the image without generating content. Behaves like Crop: one click selects the whole canvas, handles drag outward, sizes can be typed, the view zooms out so the frame stays on screen, and an attempt to make the canvas smaller answers with "use Crop".

- New `Extend` toggle next to Outpaint with its own panel: W/H fields, 1×/2×/3× per axis, aspect presets (16:9, 9:16, 4:3, 3:4, 1:1), a 3×3 "Image position" anchor grid, the shrink hint with an **Open Crop** button, Cancel/Apply. Enter applies, Escape resets.
- The frame can be dragged on the canvas (anywhere inside it, off the handles); the frame follows the pointer and the image stays put during the drag. A drag sets a custom placement (no grid cell lit); later typed sizes keep the image's offset and grow at the right/bottom.
- Default placement is top-left (canvas grows right and down).
- `CanvasExtensionTool` is extracted from `OutpaintTool` as the shared base (state, outward-only drag math, target size, aspect, placement). Outpaint keeps its arrows, centred default and pointer behaviour, pinned by `OutpaintToolRegressionTests`.
- Fit mode now fits the extended frame (plus a per-tool margin) while Extend **or** Outpaint is active, so a 3× Outpaint frame no longer runs off screen. This also fixed a pre-existing defect: the fit branch wrote `Viewport.ZoomLevel`, whose setter clears fit mode, so Fit switched itself off on the first render.
- Transparent pixels render over a transparency checkerboard (`TransparencyCheckerboard`), so extended strips and removed backgrounds are visible instead of vanishing into the canvas background. The preview strips use the same checkerboard with a green tint.
- Apply is all-or-nothing: the working bitmap is allocated first, `LayerStack.ResizeCanvas` allocates every layer's replacement before swapping any in, and a failure (Skia returns an *empty* bitmap rather than throwing) leaves the document untouched and reports a status message.
- The editor's status messages are forwarded into the shared status bar (they had no binding of their own).

Design: `docs/superpowers/specs/2026-09-02-canvas-extend-design.md`. Mockups: https://claude.ai/code/artifact/3e066a50-4879-422d-b184-58d37aa57e17 (Row A).

## Test plan

- [x] `CanvasExtendToolTests`, `OutpaintToolRegressionTests`, `ViewportFitTests`, `ImageEditorCoreCanvasExtendTests`, `CanvasExtendViewModelTests`, `ToolManagerTests` addition
- [x] Full `DiffusionNexus.Tests` suite green (2 pre-existing offline canary skips)
- [x] Release build 0 errors
- [x] Manual smoke rounds 1–4 (fields, placement, 1×, checkerboard, frame drag)
- [x] Manual smoke after the review fixes: status bar shows "Canvas extended to …" and the failure message; Fit stays on after loading an image whose fit scale is a fractional percent (window resize still re-fits); all nine anchor cells move the image and no cell is lit after dragging the frame; middle-drag pan right after load and mid-extend does not jump; typed H below the image on an extended axis snaps the frame back

## Known gaps (deferred, not regressions)

- Loading a new image while a tool panel is open does not close the panel (Outpaint has the same gap).
- A fit scale outside the viewport's zoom range (0.1–10×) renders at the true fit scale while the stored zoom is clamped, so leaving fit mode on such an image jumps. Rendering at the clamped scale would stop Fit from fitting huge images, so this is left as is.
- The `IsFitMode` setter emits a redundant `ZoomChanged` per middle press.
- A multiplier preset is a no-op when that axis already equals the request.
- `DrawResolutionLabel` duplicates `CropTool`'s label pill.
- Cursors are allocated per pointer move, as the rest of the control does; Avalonia caches the platform cursor behind them.


Little-God1983:
## Code review — Canvas Extend

Reviewed the whole diff against `develop` (24 files, 20 commits). The design is good: extracting `CanvasExtensionTool` out of `OutpaintTool` was the right call, the hit-test math and the symmetric-distribution behaviour are carried over faithfully (`OutpaintToolRegressionTests` pins them), and the test suite is unusually thorough for UI work — `SetTargetSize_OnlyResplitsTheAxisThatChanged`, `DraggingTheFrame_StopsWhenAnEdgeReachesTheImage` and `Reset_RevertsACustomPlacement_ButKeepsAChosenAnchor` all pin behaviour that would otherwise be smoke-only.

Findings below, most severe first. Nothing here breaks the feature's happy path; two of them are fixes that don't currently do what their commit messages claim.

---

### 1. A partial layer resize leaves the layer stack corrupt (and there's no undo to escape it)

`Layer.ResizeCanvas` now throws when Skia hands back an empty bitmap. `LayerStack.ResizeCanvas` calls it in a loop:

```csharp
foreach (var layer in _layers)
    layer.ResizeCanvas(newWidth, newHeight, offsetX, offsetY);

// Recorded only once every layer is resized: a failed allocation throws out of the
// loop, and the stack must not claim a size its bitmaps never got.
_width = newWidth;
_height = newHeight;
```

Moving the `_width`/`_height` assignment below the loop protects the *recorded* size, but not the bitmaps. If layer 0 allocates and layer 1 does not, layer 0 is already at the new size while the stack still reports the old one. `LayerCompositor.CompositeToCanvas` scales by `layers.Width`/`Height` and draws every layer bitmap at `0,0` unscaled — so the already-grown layers render cropped to the old canvas rect. `ApplyCanvasExtend` catches, returns `false`, the user is told "try a smaller size", and the editor carries on with a mixed-size stack. There is no undo in this editor.

**Repro:** layer mode, 2+ layers, extend to a size that allocates for the first layer and not the second (two large layers near the machine's memory ceiling).

`ImageEditorCoreCanvasExtendTests` can't see this — it never enters layer mode, so `AbsurdExtension_ReturnsFalse_AndLeavesCanvasAndToolUntouched` only exercises the working-bitmap branch.

Suggestion: make it two-phase — allocate every replacement bitmap into a local list first, then swap them into the layers (and set `_width`/`_height`) only once all allocations succeeded. Then the failure path really is a no-op, which is what the rest of the hardening in this PR assumes.

### 2. Every status message this PR adds is invisible

`ImageEditView.axaml` never binds `ImageEditorViewModel.StatusMessage`. The bottom status bar binds `ImageDimensions`, `ImageDpi`, `FileSizeText` and the zoom controls, and nothing else in any `.axaml` binds that property. So all three of these go nowhere:

- `ImageEditView.axaml.cs:583` — `"The canvas could not be extended. Try a smaller size."`
- `CanvasExtendViewModel.OnApplied` — `"Canvas extended to W x H"`
- the `IsPanelOpen` setter — `"Extend: Drag a handle outward or type a size. Press Enter to apply, Escape to reset."`

Commit `ea94b621` ("report a failed canvas extend instead of failing silently") therefore still fails silently as far as the user is concerned: on a failed apply nothing happens and nothing is said. The `CanvasExtendFailed` event, `OnApplyFailed` and the status string are all wired correctly — they just terminate at a write-only property.

There's already a pattern for the fix one level up: `ImageEditTabViewModel.OnImageActionsPropertyChanged` forwards `ImageActionsViewModel.StatusMessage` into the shared `IDatasetState.StatusMessage` ("Surface the shared actions' status … in the editor's status bar"). `OnImageEditorPropertyChanged` is right next to it and already watches `HasImage`/`CurrentImagePath`; forwarding `StatusMessage` the same way lights all of this up. Pre-existing for Outpainting/Inpainting too, so it's a small fix with wide payoff.

### 3. Fit mode still turns itself off on the first render — one hop further out

`bf2088d2` correctly identified that writing `Viewport.ZoomLevel` clears `IsFitMode`, and routes through `SetFitModeWithZoom` instead. But the zoom slider closes the loop:

`RenderWithZoom` fit branch → `viewport.SetFitModeWithZoom(clampedFit)` → `Viewport.Changed` → `ImageEditorViewModel.ZoomPercentage = 43` (ctor, ~line 388) → `Slider Value="{Binding ImageEditor.ZoomPercentage}"` (Avalonia's `RangeBase.ValueProperty` is TwoWay by default) → `WireZoomSlider`'s `onSliderChanged`, which fires on *any* `Slider.Value` change including one pushed by the binding → `SetZoom(43/100f)` → `_editorCore.ZoomLevel = 0.43f` → `ViewportManager.ZoomLevel` setter → `_isFitMode = false`.

The setter's `if (Math.Abs(_zoomLevel - clamped) < 0.0001f) return;` guard is the only thing that can save it, and it only holds when the fit scale lands on an exact integer percent. `RenderWithZoom_InFitMode_KeepsFitModeOn` uses a 100×100 image in a 400×400 surface — exactly 400% — so it passes either way; same for `..._ZoomsOutWhenTheExtendToolGrowsTheFrame` (168%). A real 1024×1024 in a ~900×700 viewport gives a fractional percent and fit mode dies on the first frame.

The symptom is subtle (renders at 43% instead of 42.71%, visually identical) but "Fit" is silently lost, so a window resize no longer re-fits — and the extend tool's zoom-out rule depends on fit mode being alive. Worth a guard in `onSliderChanged` against echoing programmatic changes, plus a test at a non-integer fit scale.

Note this also re-enters the render pass: `SetZoom` calls `InvalidateVisual()` while `RenderWithZoom` is still on the stack holding `_bitmapLock`.

### 4. Two different checkerboards — the preview doesn't match the result

`CanvasExtendTool` has its own `CheckerTile`: 16 px cells, `#2B2B2B`/`#3B3B3B`. `TransparencyCheckerboard` (added later, in `83b39e37`) is 8 px, `#666`/`#999`. Both are on screen in the same session — the tool's paints the strips *before* Apply, `TransparencyCheckerboard` paints them *after* — so the new area visibly changes appearance the instant Apply lands. That undercuts the point of previewing it as a checkerboard at all.

`TransparencyCheckerboard.Draw` also anchors its shader with `SKMatrix.CreateTranslation(rect.Left, rect.Top)`, for exactly the reason its doc comment gives; `DrawNewArea` has no matrix, so the preview pattern slides underneath the strips while panning and zooming.

`DrawNewArea` should call `TransparencyCheckerboard.Draw(canvas, strip)` per strip and keep the green tint on top; `CheckerTile`/`BuildCheckerTile` can then go.

### 5. Typing a below-image size desyncs the field from the frame

`CanvasExtendViewModel.SetTarget`:

```csharp
if (value < minimum)
{
    field = minimum;
    OnPropertyChanged(propertyName);
    IsShrinkHintVisible = true;
    return;      // no TargetSizeRequested
}
```

If that axis is already extended — say H is at 1500 on a 1024-tall image — typing `500` sets the H field to 1024 and tells the tool nothing. The frame stays at 1500, `ResolutionText` stays at "… x 1500", and nothing ever re-syncs the field (no region change → no `UpdateResolution`). The panel now contradicts the canvas.

Two defensible fixes: snap the field back to the tool's *current* total rather than to the image size, or forward `minimum` to the tool — `SetTargetSize` already handles a total equal to the image size by clearing that axis without raising `ShrinkAttempted` (`SetTargetSize_BackToTheImageSize_ClearsThatAxis_WithoutAShrinkAttempt`). `TypedWidth_BelowImage_ClampsShowsHint_AndRaisesNothing` currently pins the desync, so that test moves with the fix.

### 6. The fit rect is built from the unclamped scale, the viewport stores the clamped one

`FitImageRect` returns a rect from the raw `scale`; `SetFitModeWithZoom` clamps to `[0.1, 10]`. Outside that range (a very large image, or a 16 px icon in an 800 px canvas → fit 50×) the render and the reported `ZoomPercentage` disagree, and leaving fit mode then jumps the view — the opposite of what the middle-press comment asserts ("The stored zoom is already the clamped fit scale, so the view does not jump").

Separately, that comment doesn't hold *whenever an extension is active*, clamp or no clamp: the fit branch offsets the image inside the frame (`frameX + extendLeft * scale`) while the non-fit branch centres the image with pan 0. Middle-dragging mid-extend snaps the image to centre first. Same for Outpaint, now that it participates in `ActiveExtensionTool`.

### 7. `new Cursor(...)` on every pointer move

`CursorForExtensionHandle` allocates a fresh `Cursor` — an `IDisposable` over a native handle — per `OnPointerMoved`, as does the `IsMovePoint` branch. Pre-existing in the Outpaint code, but extracting the shared helper is the natural moment to cache nine statics.

### Nits

- `ToolManagerTests.cs` flips the line terminator on two untouched lines (line 1, and the closing `#endregion`), which is why they show up in the diff. Worth reverting so that file is a clean +11.
- 3,056 of the 6,191 added lines are the plan + spec docs (`plans/2026-09-02-canvas-extend.md` alone is 2,742). Fine if that's the intent — worth a conscious call, since the SDD ledger was deleted and these are the surviving process artifacts.
- PR body is empty. The deferred gaps belong here: loading a new image doesn't close an open tool panel (Outpaint too); the `IsFitMode` setter emits a redundant `ZoomChanged` per middle press; a multiplier preset is a no-op when that axis total already equals the request; `DrawResolutionLabel` duplicates `CropTool`'s pill.

### Smoke items this review can't cover

Findings 2 and 3 are runtime-only, and so is one more thing worth watching: the 3×3 anchor grid puts `Setter Property="Command" Value="{Binding …SetAnchorCommand}"` inside `UniformGrid.Styles`, and this project sets `AvaloniaUseCompiledBindingsByDefault=true`. It compiles, so `x:DataType` resolves from the enclosing root — but a compiled binding in a style setter resolving against the styled element's DataContext is exactly the kind of thing that only proves itself when the buttons are clicked. Verify all nine cells actually move the image, and that no cell is lit after dragging the frame.

Issue #518, opened 2026-08-17
Diffusion Canvas: turn it into a generation suite (bounding box, layers, staging, queue)


Little-God1983:
## Mockup

**[Latent Canvas — full-surface mockup](https://claude.ai/code/artifact/30275e6d-c08c-40a8-ba0e-8bf35224aa75)**

Every region below is lettered A–G on that page, with the reasoning next to it, three close-up
insets (layer row, staging bar, capability gating) and the same build map that appears at the end
of this issue. Read it first — this text is the checklist, the artifact is the design.

> Note: Claude artifacts are private until shared. If you want anyone else on this issue to open it,
> share it from the page's share menu.

---

## Why

The Canvas generates. It is not yet a suite.

The concrete symptom is that a result appears at the origin, top-left, and that is the whole spatial
model — there is nowhere to put an image, nothing to put it next to, and no way to say "redo this
part". Underneath the symptom, most of `DiffusionCanvasViewModel` is placeholders wired to disabled
controls: `NegativePromptText`, `Seed`, `Steps`, `Cfg`, `SelectedSampler`, `Loras`,
`AddControlNetCommand`, the three mask tools, `ToggleLayerPanelCommand`, `UndoCommand`,
`RedoCommand`, and a `CancelCommand` whose `CanExecute` returns `false`.

This issue replaces the spatial model. Nearly everything else on the list falls out of that change
rather than needing to be designed separately.

## The model

**An infinite canvas holding every accepted result, and one movable bounding box that declares the
generation region.**

- The box's **size** is the latent size.
- The box's **position** is where the pixels land.
- What is **under** the box is what the model gets to see.

That last point is what makes it worth building. Drag the box half off an existing result and
outpainting costs nothing extra. Drag it fully over one and it is img2img. Paint a mask inside it
and it is inpainting. One control, three features, no new modes.

### A — Backend & budget (title bar)

Active backend is a permanent segmented control in the title bar, with live VRAM/RAM beside it and
the loaded workload named.

Core and Engine do not have the same abilities, so the backend decides what the rest of the window
can offer; if it can change silently, controls stop working silently. VRAM sits next to it because
choosing a model is a VRAM decision.

- [ ] Move backend selection out of the panel into the title bar
- [ ] Dock `ResourceMonitorViewModel` there
- [ ] Add capability flags to `IDiffusionBackend` (see *Capability gating* below)

### B — Generate panel (left)

[...]

PR #558, opened 2026-09-04
Diffusion Canvas: bounding box, staging and a real cancel (#518 slice 1)

Little-God1983:
Implements the first slice of #518 — regions **C** (canvas & bounding box), **E** (staging strip) and the cancel half of **F**. Regions A, B, D and G follow on their own branches.

## What changed

The canvas's spatial model is replaced. Generate used to append a frame at a walking 40 px diagonal offset, and that was the whole story — there was nowhere to put an image and no way to say "redo this part". There is now **one movable, resizable bounding box** over an unbounded world of accepted results: its size is the latent size, its position is where the pixels land, and whatever is underneath it is what the model sees. Drag it onto a result and the run is img2img; drag it onto empty canvas and it is text2img. One control, no new modes.

Results arrive in a **staging strip** as candidates and only reach the canvas on Accept. Cancel really stops a batch.

## Notable decisions

**`ZoomBorder` is gone.** Not preference — its `Matrix`/`ZoomX`/`OffsetX` are read-only so fit and 1:1 can only be driven through a foreign API, it transforms via `RenderTransform` so every handle and border scales with zoom (the old resize grip already had this defect), its two-finger rotation gesture is on by default and a rotated world silently breaks every axis-aligned hit-test with no affordance to reset, and its zoom/pan bounds default to ±∞. The replacement follows the shape `ImageEditorControl` already uses: a custom control plus a plain transform object as the single source of truth — which also makes the whole pan/zoom/box contract unit-testable, since `DiffusionNexus.Tests` initialises no Avalonia platform.

**The region is composited arithmetically, not captured.** There is no `RenderTargetBitmap` or `SKSurface` anywhere in this repo and no precedent for snapshotting a live Avalonia visual tree, so `CanvasRegionCompositor` draws the accepted rasters' own PNGs into a bitmap at the box's latent size. That keeps the result independent of zoom and DPI scaling, and it reports an opaque-coverage fraction so the caller can pick text2img or img2img.

**img2img works on both backends.** The local backend already honoured `InitImage` (its "Currently ignored by the v1 backend" doc was stale). For the Engine, the shipped `Krea2-Text2Image-API.json` already carries a `VAELoader`, so passing an uploaded region injects `LoadImage → VAEEncode` and repoints `KSampler.latent_image` — both core ComfyUI types, so no new template asset and nothing added to the required-custom-nodes list.

**Status + VRAM monitor have their own row.** The monitor is ~650 logical px wide and the window's minimum width is ~940, so sharing a row with the tool strip left the strip ~240 px. This is the title bar region A describes, brought forward.

**Batch width is 1, deliberately.** `DiffusionContextHost` holds a per-model `SemaphoreSlim(1,1)` with a single-resident policy, so parallel canvas generations would either serialise behind that lock or thrash VRAM.

## Honest limits

- **Partial coverage is not outpainting.** With no `MaskImage` support in either backend, the uncovered part of a partly-covered box is flattened onto neutral grey and the whole region is denoised — the known pixels are *regenerated*, not preserved. The status line reports the coverage percentage and the log warns. Masked outpainting belongs with the layer stack (region D).
- **Local cancel cannot interrupt mid-sample.** `stable-diffusion.cpp` exposes no cancel hook, so Cancel drops the queued remainder and discards the in-flight image once the native call returns. The Engine *is* interrupted immediately via `POST /interrupt`. The tooltip says which is which.
- The Engine re-runs its full readiness probe once per image in a batch. Left as a TODO naming the cost: memoising it needs an invalidation that cannot mask a genuinely stopped engine.

## Review

A 7-dimension adversarial pass (3 independent refuters per finding, majority kills) produced 57 raw findings; 26 survived and are fixed in `dad4a147`. The ones worth knowing about: Cancel was completely inert for the whole pre-flight phase (up to two minutes on a cold engine start); the engine interrupt was fire-and-forget and raced by the disposal of the very `HttpClient` sending it; cancelling a local batch orphaned the producer while it still held the model lock, so the UI said Idle while the next Generate silently blocked; and Alt-dragging could leave a box size off the model's lattice that made Generate refuse every subsequent click.

A second, independent review of the PR (15 inline findings) is addressed in `77de77fb`: the server-global `/interrupt` fired on pre-flight cancels, backend timeouts reported as user cancels, the lost right-click Delete, the readout promising img2img for a run that would execute as text2img, the per-launch leak into the engine's `input/` folder, prompt provenance read at accept time, two alignment edge cases, Escape leaving Alt snapping in effect, the 20 fps full-surface repaint, per-move cursor allocation, the unused `WorldRect`, the per-candidate init-image re-upload, a test that could not fail, and the CRLF flip on six files. The one part pushed back on is adding `.gitattributes` in this PR: with 446 CRLF blobs in the index it needs a renormalise commit of its own.

## Verification

- `dotnet build DiffusionNexus.sln -c Debug` — 0 errors (12 pre-existing CA1416 warnings).
- `dotnet test DiffusionNexus.Tests` — **5384 passed**, 0 failed (up from 5305 on `develop`).
- **GUI smoke, done by hand on both backends (Core and Engine):** prompt round-trip, box move/resize, text2img over empty canvas, img2img over an accepted result, multi-image batches, ←/→ stepping, Space-to-compare, cancel mid-batch. Three defects found and fixed during the smoke: the tool strip painting over the VRAM monitor (`914d77f3`, `b5e25463`) and cancelled slots lingering as blank tiles in the staging strip (`f7ba5259`).
- Layout verified at the window's minimum width and at 2600 px via a captured Debug instance.


DiffusionNexus.UI/Services/Diffusion/ManagedComfyUiBackend.cs
```diff
@@ -259,6 +271,15 @@ public async IAsyncEnumerable<DiffusionStreamItem> GenerateAsync(
...
         {
+            // A cancelled canvas batch must stop work on the engine, not merely stop waiting for it —
+            // otherwise the sampler runs to completion server-side and burns the GPU on a result nobody
+            // reads. Issued here, on the unwind, rather than from a CancellationToken.Register callback:
+            // that callback can only fire-and-forget, and the same cancellation tears down this method's
+            // `using var wrapper` moments later. Disposing the wrapper disposes the HttpClient it owns,
+            // which aborts every request still in flight — including the interrupt that was supposed to
+            // stop the GPU. Awaiting it here keeps the client alive until the POST lands.
+            // InterruptAsync swallows its own failures, so this cannot mask the cancellation.
+            await wrapper.InterruptAsync(CancellationToken.None).ConfigureAwait(false);
```
Little-God1983: **`/interrupt` is global, and it is fired even when this call never queued anything.**

The catch filter is `OperationCanceledException when (cancellationToken.IsCancellationRequested)`, which covers the whole `try` — including `ResolveInstalledKreaGgufAsync`, `LoadTemplateJson` and `UploadInitImageAsync`, all of which run *before* `QueueWorkflowAsync`. ComfyUI's `/interrupt` is not prompt-scoped: it abandons whatever prompt the server is executing right now.

Failure scenario: an Inpainting / Outpainting / Batch-Upscale job (which all talk to a ComfyUI through `ComfyUIWrapperService`) is mid-sample on the same server; the user starts a canvas batch and hits Cancel during pre-flight (up to ~2 minutes on a cold engine, and Cancel is deliberately live for all of it). This canvas run had queued nothing, but the other feature's prompt is torn down server-side and comes back with no image.

Gate the interrupt on having actually queued: hoist `promptId` to a `string?` declared before the `try` and only send the interrupt when it is non-null.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -1,572 +1,1122 @@
...
+                ? null
+                : new DiffusionReferenceImage(regionImagePath, (float)DenoiseStrength);
+
+            var candidates = Staging.BeginBatch(BatchCount, region);
+            EmitInfo($"Staged {candidates.Count} candidate slot(s).");
+
+            await RunBatchAsync(backend, descriptor, candidates, initImage, width, height, coverage, token)
+                .ConfigureAwait(true);
+        }
+        catch (OperationCanceledException)
```
Little-God1983: **An unfiltered `catch (OperationCanceledException)` reports a backend timeout as a user cancel.**

Nothing here checks `token.IsCancellationRequested`, so *any* `OperationCanceledException` that escapes the pre-flight is reported as `"Cancelled."` and logged as `"Batch cancelled by the user."`. `HttpClient` raises `TaskCanceledException` (an `OperationCanceledException`) on its own timeout, and `ResolveBackendAsync` → `ManagedComfyUiBackend.IsAvailableAsync` → `CheckRequiredNodesAsync` issues `GET /object_info` on a client whose timeout is 10 minutes (`ComfyUIWrapperService.CreateDefaultHttpClient`).

Failure scenario: the engine process is alive but wedged (a custom node hanging import, a stuck model load). After 10 minutes the readiness probe throws `TaskCanceledException`; the canvas prints "Cancelled.", prunes the strip and writes "Batch cancelled by the user." to the Unified Console. The user never clicked Cancel and gets no diagnostic at all — the one place the real cause would have shown up is swallowed.

`ManagedComfyUiBackend` already gets this right with `when (cancellationToken.IsCancellationRequested)`; this catch needs the same filter so a genuine timeout falls through to the `catch (Exception ex)` error path.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -1,572 +1,1122 @@
...
+            CanvasY = candidate.WorldRect.Y,
+            Width = (int)Math.Round(candidate.WorldRect.Width),
+            Height = (int)Math.Round(candidate.WorldRect.Height),
+            Prompt = PromptText,
+            Seed = candidate.Seed,
+            FrameImage = image,
+            ImagePath = path,
+            State = GenerationFrameState.Completed,
+            StatusText = candidate.StatusText,
+            DeleteCommand = DeleteFrameCommand,
```
Little-God1983: **Per-frame delete was dropped in the rewrite, but its wiring was kept — so there is now no way to remove a single result.**

`DeleteFrameCommand` is constructed in all three constructors and handed to every frame here, yet nothing binds it any more. The old view had a per-frame `ContextMenu` with `MenuItem Header="Delete frame"` (`develop:DiffusionCanvasView.axaml:151-164`); the new `DiffusionCanvasView.axaml` has no `ContextMenu`/`ContextFlyout`, and `DiffusionCanvasSurface` is a custom `Control` that ignores right-click entirely (`OnPointerPressed` returns unless `IsLeftButtonPressed`).

Failure scenario: the user accepts five candidates, decides the third is bad, and right-clicks it — nothing happens. The only remaining removal path is **Clear canvas**, which wipes all five, and `UndoCommand` is an `AlwaysFalse` placeholder, so that is unrecoverable.

Either restore a right-click affordance on the surface (hit-test the raster under the pointer and open a flyout on `DeleteFrameCommand`) or delete `DeleteFrameCommand`, `DeleteFrame` and `GenerationFrameViewModel.DeleteCommand` so the dead wiring does not read as a working feature.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -1,572 +1,1122 @@
...
+    // ────────────────────────────── Region under the box ──────────────────────────────
+
+    /// <summary>
+    /// Recomputes whether the box currently overlaps any accepted result. Cheap — it is rectangle
+    /// intersection only, no pixel work — so it can run on every box move.
+    /// </summary>
+    private void RefreshRegionMode()
+    {
+        var region = Box.WorldRect;
+        var overlapping = Frames.Count(f => new Rect(f.CanvasX, f.CanvasY, f.Width, f.Height).Intersects(region));
```
Little-God1983: **`RegionModeText` / `IsRegionOccupied` are decided by rectangle intersection alone, so the UI can promise img2img on a run that silently executes as text2img.**

`BuildRegionInitImage` returns `(null, 0)` in two cases this predicate cannot see: every intersecting raster is skipped by `LoadIntersecting` (null/missing `ImagePath`, undecodable PNG), or `composite.IsEmpty` (no opaque pixels under the box). Both paths only `EmitInfo` — the status line and the denoise slider keep saying image-to-image.

Failure scenario: `OnCandidateAccepted`'s `OutputsWriter` throws (outputs folder on a full or read-only drive) — the exception is caught and the frame is added with `ImagePath = null`. The result is visible on the canvas, the box sits squarely on it, the readout says *"Image to image — the box is over 1 result"* and the Denoise slider is shown, but `LoadIntersecting` skips it and the backend receives `InitImage = null`. The user sets denoise to 0.2 expecting a light retouch and gets an unrelated fresh image, with the only explanation buried in the Unified Console.

Count only rasters that could actually contribute (non-empty, existing `ImagePath`) so the readout matches the run.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -1,572 +1,1122 @@
...
+    /// Scratch file the composited region is written to before a backend reads it.
+    /// </summary>
+    /// <remarks>
+    /// Deliberately stable for the life of the view model rather than a fresh GUID per run. The engine
+    /// backend uploads this file into ComfyUI's own <c>input/</c> folder, and that copy is never deleted:
+    /// a per-run name left hundreds of multi-megabyte PNGs behind over a session. A stable name means the
+    /// upload overwrites in place (<c>UploadImageAsync</c> already posts <c>overwrite=true</c>). The GUID
+    /// still distinguishes concurrent app instances.
+    /// </remarks>
+    private readonly string _regionScratchPath =
```
Little-God1983: **The stable scratch name leaks one multi-megabyte PNG into ComfyUI's `input/` per app launch, forever.**

The `Guid.NewGuid()` is evaluated once per view-model instance, and this view model is a DI singleton — so the name is stable *within* a run but different on every launch. `ManagedComfyUiBackend.UploadInitImageAsync` copies the file into the engine's own `input/` folder and nothing ever deletes it (`DeleteScratchFile` only removes the *temp* copy). The XML remark's reasoning — "a per-run name left hundreds of multi-megabyte PNGs behind over a session" — holds for a session but the same unbounded growth just moves to a per-launch cadence.

Failure scenario: a user who opens the app daily and uses the canvas accumulates one 2048×2048 PNG (~5–10 MB) per launch in the engine's `input/`, with no UI that lists or clears it; after a few months that is a silent multi-gigabyte directory.

Drop the GUID for a fixed name (e.g. `diffusionnexus_canvas_region.png`) so `overwrite=true` genuinely overwrites in place across launches, or have the engine backend clean up the name it uploaded.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -1,572 +1,1122 @@
...
+        var image = candidate.Image;
+        candidate.Image = null;
+
+        var frame = new GenerationFrameViewModel
+        {
+            CanvasX = candidate.WorldRect.X,
+            CanvasY = candidate.WorldRect.Y,
+            Width = (int)Math.Round(candidate.WorldRect.Width),
+            Height = (int)Math.Round(candidate.WorldRect.Height),
+            Prompt = PromptText,
```
Little-God1983: **The accepted frame records the prompt as it is *now*, not the prompt that produced the image.**

`StagedCandidateViewModel` snapshots `WorldRect` at `BeginBatch` precisely so a later box move cannot retarget a finished result, but the prompt gets no such snapshot — `PromptText` is read here, at accept time.

Failure scenario: the user generates four candidates, then starts typing the next idea into the prompt box while judging them, then presses Enter to accept. `frame.Prompt` stores the half-typed *next* prompt. The field is documented as "kept for provenance", and it is the only record of what made the image once the strip is gone — an accepted result that region D / the planned context menu will later surface (Copy prompt) now carries the wrong text.

Capture the prompt onto the candidate in `RunBatchAsync` (next to `Seed`) and read it back here.

DiffusionNexus.UI/DiffusionCanvas/GenerationBoundingBox.cs
```diff
@@ -0,0 +1,302 @@
...
+        // snapped value back onto the lattice if rounding pushed it outside the bounds (possible when
+        // the alignment does not divide MinSize/MaxSize).
+        var clamped = Math.Clamp(desired, MinSize, MaxSize);
+        var snapped = (int)Math.Round(clamped / _alignment) * _alignment;
+        if (snapped > MaxSize)
+            snapped -= _alignment;
+        if (snapped < MinSize)
+            snapped += _alignment;
+
+        return Math.Clamp(snapped, _alignment, MaxSize);
```
Little-God1983: **Two arithmetic edge cases around the model's alignment: this `Math.Clamp` can throw, and the caller divides by the unclamped value.**

1. `Math.Clamp(int, min, max)` throws `ArgumentException` when `min > max`. Here `min` is `_alignment` and `max` is `MaxSize` (2048), so any `Alignment` above 2048 makes every `SnapSize` call throw — including the one inside the `Alignment` setter itself.
2. The setter defends the field with `Math.Max(1, value)`, but `DiffusionCanvasViewModel.GenerateAsync` line 549 immediately does `Box.Width % descriptor.DimensionAlignment` against the **raw** descriptor value. A descriptor with `DimensionAlignment = 0` therefore survives the setter and then raises `DivideByZeroException` one line later — the guard and its consumer disagree about what the alignment is.

Failure scenario: a catalog entry ships `DimensionAlignment = 0` (the property is `init` with a default, so an explicit `0` is one typo away). Generate dies with a bare `DivideByZeroException` surfaced as `"Error: Attempted to divide by zero."` instead of a fixable message.

Clamp with `Math.Clamp(snapped, Math.Min(_alignment, MaxSize), MaxSize)` (or validate in the `Alignment` setter), and read the sanitised value back from `Box.Alignment` at line 549 rather than re-using the descriptor's raw field.

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
```diff
@@ -0,0 +1,680 @@
...
+            held = false;
+        if (IsSpaceHeld == held)
+            return;
+
+        IsSpaceHeld = held;
+        Cursor = held ? new Cursor(StandardCursorType.Hand) : Cursor.Default;
+    }
+
+    /// <summary>Abandons an in-progress box gesture, restoring the box to where the drag began.</summary>
+    public void CancelActiveGesture()
```
Little-God1983: **Escape leaves `SnapPositionToGrid` stuck off — the other two gesture-end paths restore it, this one does not.**

`OnPointerMoved` sets `Box.SnapPositionToGrid = !Alt` on every move of a box drag. `OnPointerReleased` and `OnPointerCaptureLost` both explicitly restore it to `true` "regardless of how this one ended"; `CancelActiveGesture` — the Escape path documented in `Shortcuts.md` as *"Abandon the in-progress box drag"* — calls `Box?.CancelDrag()` and `ReleaseGesture()` and skips the restore.

Failure scenario: Alt-drag the box, press Escape. `_isDraggingBox` is now false, so the still-held button's `PointerReleased` skips its restore branch too, and the box is left with `SnapPositionToGrid == false`. Any position change that is not a pointer drag — `CenterOn`, and every future caller of `SetPosition` — then lands off-lattice, which is the state commit `dad4a147` set out to make impossible.

Add `box.SnapPositionToGrid = true;` next to the `CancelDrag()` call.

DiffusionNexus.Service/Services/ComfyUIWrapperService.cs
```diff
@@ -600,6 +600,33 @@ public async Task<IReadOnlyList<string>> GetNodeInputOptionsAsync(
         return [];
     }
 
+    /// <inheritdoc />
+    public async Task InterruptAsync(CancellationToken ct = default)
+    {
+        ObjectDisposedException.ThrowIf(_disposed, this);
```
Little-God1983: **This throws before the try that is documented to swallow everything, and it is called from inside a catch block.**

The interface contract added in this PR says *"Failures are swallowed by design. This is called from a cancellation path, where the engine may already be stopping, unreachable, or between prompts; throwing there would replace a clean cancellation with an unrelated error."* — but `ObjectDisposedException.ThrowIf` sits above the `try`, so a disposed wrapper does exactly what that remark forbids.

Failure scenario: any caller that interrupts on a teardown path where the wrapper has already been disposed (a shutdown-ordering change, or a future caller holding a shared wrapper rather than `ManagedComfyUiBackend`'s method-local `using var`). The `ObjectDisposedException` is thrown from *inside* `catch (OperationCanceledException) { await wrapper.InterruptAsync(...); throw; }`, so it replaces the cancellation and surfaces as a generation failure.

Move the disposed check inside the `try`, or return early instead of throwing.

DiffusionNexus.Inference/StableDiffusionCpp/StableDiffusionCppBackend.cs
```diff
@@ -1,351 +1,373 @@
...
-                NativeLog.Debug("{NativeText}", text);
-                break;
-        }
-    }
-
-    public void Dispose() => _host.Dispose();
-
-    private readonly record struct GenerationOutcome(byte[] PngBytes, int Width, int Height, long Seed, int TotalSteps);
-}
+using System.Linq;
```
Little-God1983: **This file's line endings were flipped CRLF to LF, so a 24-line change reads as a 724-line rewrite.**

`git show develop:<path>` reports CRLF, `git show HEAD:<path>` reports LF. The only real change here is the `try`/`finally` around the `await foreach` (+24/-2 under `git diff -w`), but every line shows as touched, which buries the actual edit and will conflict with anything else that lands on this file.

Six files in the PR did this: `StableDiffusionCppBackend.cs`, `DiffusionCanvasViewModel.cs`, `GenerationFrameViewModel.cs`, `DiffusionCanvasView.axaml`, `DiffusionCanvasView.axaml.cs`, `Doc/Shortcuts.md`. The repo has no `.gitattributes`, so nothing normalises this — the same class of problem `Installer.SDK` PR #57 had to fix. Worth restoring CRLF on the touched files, and separately adding `* text=auto` so it stops recurring.

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
```diff
@@ -0,0 +1,680 @@
...
+        else if (change.Property == ShowGridProperty
+              || change.Property == PreviewImageProperty
+              || change.Property == PreviewRectProperty
+              || change.Property == IsPreviewHiddenProperty)
+        {
+            InvalidateVisual();
+        }
+    }
+
+    private void OnAntsTick(object? sender, EventArgs e)
```
Little-God1983: **The marching-ants timer forces a full-surface repaint at 20 fps for as long as the tab is open, whether or not anything is happening.**

`OnAntsTick` only checks `Box is null`, and `Box` is bound to a view-model property that is never null in production, so `InvalidateVisual()` fires 20 times a second permanently. Each of those runs the whole `Render`: `DrawGrid` issues one `FillRectangle` per lattice dot (~500 at zoom 1 on a 1080p surface, ~800 at min zoom), then `DrawRasters` re-issues a `DrawImage` for every accepted result, then `DrawPreview`, then `DrawBox` — which additionally allocates a fresh `Pen` and `DashStyle` each frame.

Cost: on the order of 10-16k rectangle fills per second plus every raster re-drawn, purely to animate a dashed border, on an idle screen. On a laptop that is continuous CPU/GPU draw any time the Diffusion Canvas tab is the visible one.

Cheapest fix that keeps the affordance: draw the ants in a small overlay `Control` layered above the surface so only its box-sized bounds are invalidated, and hoist the ants `Pen`/`DashStyle` out of `DrawBox` into cached fields keyed by offset.

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
```diff
@@ -0,0 +1,680 @@
...
+            return;
+
+        if (Box is not { } box)
+        {
+            Cursor = Cursor.Default;
+            return;
+        }
+
+        var handle = box.HitTest(Viewport.ScreenToWorld(screen), Viewport.ScreenToWorldLength(HandleHitScreenRadius));
+        Cursor = new Cursor(handle switch
```
Little-God1983: **A new `Cursor` is allocated on every pointer-move.**

`UpdateCursor` runs from `OnPointerMoved` for every move event that is not a pan or a box drag, and unconditionally constructs `new Cursor(...)` — including for `StandardCursorType.Arrow`, i.e. the common "pointer is over nothing" case. `Cursor` is `IDisposable` and none of these are disposed; every assignment also raises the styled-property change and pushes a cursor update to the platform even when the shape did not change.

Hold the seven shapes as `static readonly Cursor` fields (plus `Cursor.Default`), map the handle to one of them, and assign only when it differs from the current value. `ReleaseGesture` and `BeginPan` (line 472) allocate the same way.

DiffusionNexus.UI/DiffusionCanvas/ICanvasRaster.cs
```diff
@@ -0,0 +1,38 @@
...
+    Bitmap? FrameImage { get; }
+
+    /// <summary>
+    /// Absolute path of the saved PNG. The compositor decodes from here rather than from
+    /// <see cref="FrameImage"/>, so compositing needs no Avalonia platform and stays unit-testable.
+    /// </summary>
+    string? ImagePath { get; }
+
+    /// <summary>The raster's world rectangle.</summary>
+    Rect WorldRect => new(CanvasX, CanvasY, Width, Height);
```
Little-God1983: **`WorldRect` is added to this interface and then never used — four call sites hand-roll it instead.**

`CanvasRegionCompositor.LoadIntersecting` line 138, `DiffusionCanvasSurface.GetContentWorldBounds` line 342 and `DiffusionCanvasSurface.DrawRasters` line 581 all build `new Rect(raster.CanvasX, raster.CanvasY, raster.Width, raster.Height)` by hand, and `DiffusionCanvasViewModel.RefreshRegionMode` line 355 does the same over `GenerationFrameViewModel`. The default interface member this PR introduces is dead.

Cost: four copies of the same construction to keep in step if a raster's world footprint ever stops being exactly `(CanvasX, CanvasY, Width, Height)` — e.g. when the layer stack (region D) adds an offset or a scale. Call `raster.WorldRect` at all four sites, or drop the member.

DiffusionNexus.UI/Services/Diffusion/ManagedComfyUiBackend.cs
```diff
@@ -272,6 +293,35 @@ public async IAsyncEnumerable<DiffusionStreamItem> GenerateAsync(
...
+            return null;
+
+        if (!File.Exists(init.FilePath))
+        {
+            Logger.Warning(
+                "The init image {Path} no longer exists; generating text-to-image instead.", init.FilePath);
+            return null;
+        }
+
+        var stored = await wrapper.UploadImageAsync(init.FilePath, ct).ConfigureAwait(false);
```
Little-God1983: **The same init image is re-read and re-POSTed once per candidate in a batch.**

The canvas composites the region once per Generate and hands every candidate in the batch the identical `InitImage` path, but `UploadInitImageAsync` runs inside `GenerateAsync`, which the canvas calls once per candidate. A batch of 8 therefore performs 8 x (`File.ReadAllBytes` of the whole PNG + a multipart POST of the same bytes, to the same filename, with `overwrite=true`).

Cost: at 2048x2048 that is roughly 40-80 MB read and pushed over the loopback socket to produce a file that is byte-identical every time — on top of the full readiness probe this method already re-runs per candidate (the TODO at line 163). Both are fixed by the same seam: cache the last `(FilePath, LastWriteTimeUtc, Length) -> storedName` in the backend and skip the upload on a hit.

DiffusionNexus.Tests/DiffusionCanvas/DiffusionCanvasBatchTests.cs
```diff
@@ -0,0 +1,674 @@
...
+
+        await vm.GenerateCommand.ExecuteAsync(null);
+
+        vm.Box.Alignment.Should().Be(128);
+        vm.Box.Width.Should().Be(1024, "1088 is not a multiple of 128, so the box is re-snapped");
+        backend.LastRequest!.Width.Should().Be(1024);
+    }
+
+    [Fact]
+    public async Task Generate_NormalisesTheBoxEvenWhenTheAlignmentIsUnchanged()
```
Little-God1983: **This test cannot fail — the state it claims to cover is unreachable through the API it uses.**

The premise is *"a box that slipped off the lattice must be repaired here"*, but `SetSize` routes through `SnapSize`, which always snaps: `SetSize(1000, 1000)` stores 1024x1024 before `Generate` is ever called. `SnapPositionToGrid = false` governs position only (documented as such on the property). So the box is already aligned at the start, and the assertions `(Width % 64) == 0` / `(Height % 64) == 0` hold no matter what `GenerateAsync` does — deleting the `Box.Alignment = descriptor.DimensionAlignment` line under test would leave this green.

To actually cover the re-snap, seed a genuinely misaligned size (e.g. `Box.Alignment = 100; Box.SetSize(1000, 1000);` giving 1000, then generate against a 64-alignment descriptor) and assert the resulting *value*, not its divisibility.

DiffusionNexus.UI/Services/Diffusion/ManagedComfyUiBackend.cs
Little-God1983: Fixed in 77de77fb. `promptId` is declared above the try as `string?` and initialised null; the cancellation catch calls `InterruptAsync` only when it is non-null, so a cancel during pre-flight (GGUF lookup, template load, init-image upload) never touches whatever the server is executing for someone else.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
Little-God1983: Fixed in 77de77fb. Both catches now carry `when (token.IsCancellationRequested)`: the one in `GenerateAsync` and the per-candidate one in `RunBatchAsync`. A `TaskCanceledException` from a backend timeout during pre-flight lands in the error path ("Error: …" plus the Unified Console entry); one mid-batch fails just that candidate and the batch carries on. Tests: `ABackendTimeoutDuringPreflightIsAnErrorNotACancel`, `ABackendTimeoutMidBatchFailsThatCandidateAndTheBatchCarriesOn`.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
Little-God1983: Restored rather than removed, in 77de77fb. The surface resolves the topmost raster under a right-button press (new `CanvasRasterHitTest`, z-order tested) and on release opens a `MenuFlyout` with "Delete result", bound through a new `DeleteRasterCommand` styled property to the view model's `DeleteFrameCommand` with the raster as parameter. The dead `GenerationFrameViewModel.DeleteCommand` is gone. Documented in Shortcuts.md and the REUSABLES row.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
Little-God1983: Fixed in 77de77fb, in two halves. The readout now counts only rasters with a non-null `ImagePath`, so the OutputsWriter-failed frame in your scenario reads as text-to-image and runs as text-to-image (`AFrameWithoutASavedFileDoesNotCountAsImageToImage`). For what the readout cannot see without a disk hit per move (path present but file missing or undecodable, or an empty composite) `BuildRegionInitImage` now reports `Degraded` and Generate refuses with a status message instead of silently running text-to-image (`Generate_RefusesWhenTheResultUnderTheBoxCannotBeReadBack`). Refusing over degrade-with-warning was a deliberate choice: a run that does something other than what the UI promised is worse than no run.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
Little-God1983: Fixed in 77de77fb. The file name is now the fixed `diffusionnexus_canvas_region.png` (app-prefixed, because `overwrite=true` would clobber a user's file of the same name in input/), inside a per-process directory `%TEMP%\DiffusionNexus\canvas-<pid>\` that keeps concurrent app instances apart. The engine therefore holds exactly one such file, overwritten in place across launches; the directory is removed on Dispose. Test `Generate_WritesTheRegionUnderOneFixedFileName`.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
Little-God1983: Fixed in 77de77fb. `GenerateAsync` snapshots `PromptText` once per batch, `RunBatchAsync` stamps it on each `StagedCandidateViewModel.Prompt` and uses it for every request, and `OnCandidateAccepted` reads `candidate.Prompt`. The same snapshot closes a second hole the old per-candidate read had: a prompt typed mid-batch leaked into later candidates' requests. `AcceptedFrameRecordsThePromptTheCandidateWasGeneratedFrom` covers both.

DiffusionNexus.UI/DiffusionCanvas/GenerationBoundingBox.cs
Little-God1983: Fixed in 77de77fb. The `Alignment` setter clamps into `[1, MaxSize]`, and `GenerateAsync` reads `Box.Alignment` back for the modulo check and the messages instead of the descriptor's raw field. Tests: `AZeroAlignmentIsSanitisedRatherThanStored`, `AnAlignmentAboveTheMaximumSizeCannotMakeSnappingThrow`, `Generate_SurvivesACatalogEntryWithZeroAlignment`.

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
Little-God1983: Fixed in 77de77fb. `CancelActiveGesture` sets `SnapPositionToGrid = true` next to `CancelDrag()`, the same restore the two release paths do.

DiffusionNexus.Service/Services/ComfyUIWrapperService.cs
Little-God1983: Fixed in 77de77fb. `InterruptAsync` returns early (with a debug log line) when the wrapper is disposed instead of throwing, so it honours the never-throws contract from inside a caller's catch block.

DiffusionNexus.Inference/StableDiffusionCpp/StableDiffusionCppBackend.cs
Little-God1983: The six files are CRLF again in 77de77fb, verified with `git ls-files --eol` and `git diff --numstat develop`: StableDiffusionCppBackend is back to 27/5, GenerationFrameViewModel to 50/37, and so on. New files stay LF, which is the majority convention here (890 LF vs 446 CRLF blobs in the index).

On `* text=auto`: pushing back on doing it in this PR. With 446 CRLF blobs in the index, that attribute makes git renormalise each of them to LF the next time it is touched, which is the same whole-file diff this thread is about, spread across every future PR. It only works cleanly together with a one-shot `git add --renormalize .` commit, and that is a 446-file repo-wide change that deserves its own PR and its own decision.

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
Little-God1983: Fixed in 77de77fb. The box (ants, handles, readout) is drawn by a private `BoxLayer` child visual; the timer invalidates only that layer, so the grid, the rasters and the preview are re-recorded only when their own state changes. The ants pens are eight cached `ImmutablePen`s (one per dash offset) and the readout's `FormattedText` is rebuilt only when its string changes.

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
Little-God1983: Fixed in 77de77fb. Cursors come from a lazily filled per-shape cache (`Arrow` maps to `Cursor.Default`), and `ApplyCursor` assigns only when the instance differs, at every site including `ReleaseGesture` and `BeginPan`.

DiffusionNexus.UI/DiffusionCanvas/ICanvasRaster.cs
Little-God1983: Fixed in 77de77fb. All four sites use `WorldRect`. `GenerationFrameViewModel` implements it directly, because `Frames` is typed to the class and a default interface member is not reachable through a class reference; the compositor, the two surface sites and the new hit-test helper call it through the interface.

DiffusionNexus.UI/Services/Diffusion/ManagedComfyUiBackend.cs
Little-God1983: Fixed in 77de77fb. `ManagedComfyUiBackend` keeps an `UploadedInitImageCache`: one entry keyed on path plus a SHA-256 of the content (a timestamp is too coarse for a file that is rewritten at the same path on every Generate), and a hit is additionally gated on the stored file still existing in `<ComfyUI>/input/`, so a reinstall or a manual clean-up cannot make the graph name a file the server lacks. Six tests on the cache. The per-candidate readiness probe stays the TODO it was.

DiffusionNexus.Tests/DiffusionCanvas/DiffusionCanvasBatchTests.cs
Little-God1983: Fixed in 77de77fb. The test seeds a genuinely misaligned box by setting `Alignment = 100` and `SetSize(1000, 1000)` (asserted to hold 1000), generates against a 64-lattice, and asserts the value 1024 on both the box and the request. Deleting the `Box.Alignment = …` line under test now fails it.

PR #559, opened 2026-09-04
feat(canvas): generate panel with per-backend capability gating (#518 region B)

Little-God1983:
Implements **region B** of #518 (the generate panel), together with the capability half of **region A** and the issue's *Capability gating* section. The issue's own suggested order groups these: "Generate panel bindings + capability gating (B, A) — mostly wiring, safe to run in parallel".

Slice 1 (regions C, E, cancel half of F) shipped in #558.

## Why these ship together

Region B cannot ship truthfully on its own. The two backends disagree on nearly every parameter a panel exposes:

| Parameter | Local (stable-diffusion.cpp) | Engine (Krea 2 / ComfyUI) |
|---|---|---|
| Steps, guidance | honoured | honoured |
| Sampler, scheduler | honoured | **ignored** — baked into the template's KSampler |
| Negative prompt | **was ignored**, now wired | honoured |
| LoRAs | honoured | **ignored** — the graph's loader node is never patched |

A sampler dropdown wired naively is a per-backend lie: the user picks `dpmpp2m`, the engine silently runs `euler`, and the model gets the blame. So every control the selected backend cannot honour is **disabled carrying the backend's own one-line reason** — the rule the issue already states, that hiding teaches nothing and greying without a reason reads as a bug.

`BackendCapabilities` is deliberately not a bool set and deliberately not a default interface member. The reason has to appear at the control, and the backend is what knows why; a new backend must state its limits rather than inherit a claim to support everything.

## What the panel does

One scrolling column ordered by how often it is touched: model, prompt, negative, LoRAs, then sampling folded away with its values in the header (`euler · 9 · cfg 1.0`). Generate is pinned to the footer and never scrolls out of reach.

Backend selection moved to the title bar beside the VRAM gauges, because picking a model is a VRAM decision. Status text moved to the panel footer above Generate, which is where the user is already looking during a run.

**Selecting a model adopts its own sampling defaults.** This is correctness, not convenience: the panel sends what it shows, so keeping the previous model's numbers would silently override the new one's defaults, and FLUX.2-klein wants 20 steps where Qwen wants 4.

**Prompt length is counted in characters and never capped.** A token count needs the model's own tokenizer, and the familiar 77 limit is CLIP's, which these T5-based models do not use. A fabricated number is worse than none.

## Two defects that blocked honesty

- **The engine backend never read `request.ModelKey`.** It ran the Krea 2 graph for whatever was selected, while its catalog unions Krea 2 with a real disk scan of the engine's models root. After one readiness probe the dropdown could offer Z-Image-Turbo or FLUX.2-klein, and picking one produced a Krea 2 image reported under the other model's name. Now refused as data, before the probe.
- **The local backend dropped `NegativePrompt`** although the shipped binding exposes `WithNegativePrompt`. Gating that as "unsupported" would have recorded a TODO as a capability.

## LoRAs, and one correction to the issue

The picker, the row view model and the catalog were already reusable; the canvas simply never filled `DiffusionRequest.Loras`. Wired for the local backend with no backend change, and gated with a reason on the engine.

The issue says to filter "the same rule the LoRA Viewer's *Download to existing source folder* already applies". **That flow applies no compatibility rule at all** — it only picks a folder name. The real rule is the catalog's exact case-insensitive set membership against raw Civitai labels, and there was no mapping from a canvas model to such a label anywhere.

New `ModelBaseModelLabels` is that map, authored rather than derived: a descriptor's `DisplayName` is not the Civitai label and no transform bridges them. Krea 2 maps to an explicitly **empty** list, because Civitai has no Krea 2 base model at all. Never a null filter — the catalog reads null as "return every installed LoRA", thousands of rows each decoding a thumbnail from a database BLOB. The two pipeline view models collapse onto the same map, removing a duplicated array.

## Also fixed

- The tunnel key handler exempted only a focused `TextBox`, so the panel's sliders would have lost their arrow keys to the staging strip and its dropdowns Enter and Space.
- The dot grid was hard-coded to 64 while the box snaps to the selected model's alignment, so after one Generate on a 16-aligned model the drawn guide contradicted the actual snapping.
- Guidance and denoise labels are formatted invariantly. Through a XAML `StringFormat` they rendered "1,0" on this German-locale machine, beside a header reading "cfg 1.0".
- The batch-test fakes moved into their own file so a second canvas test file can use them.

## Deliberately out of scope

- **Persistence.** The canvas persists nothing today and a new settings column costs a migration plus a recovery-service entry. Region B does not ask for it.
- **A width/height control.** The bounding box owns dimensions; a second editor for the same number invites the two to disagree.
- **Engine LoRAs.** Needs a graph patch (small) plus an absolute-path to ComfyUI-relative-name resolver that does not exist, over two separately-populated registries. Gated with a reason instead.

## Review

An adversarial pass over the branch (six dimensions, every finding independently refuted before it survived) raised 36 findings; 29 were refuted and 6 are fixed in `dc5c6099`:

- **One click on a panel button killed every keyboard shortcut.** The guard exempted anything inside the panel, and focus lands on a Button when clicked and stays there, so clicking Unload or a seed control silently disabled the canvas and staging keys until the canvas was clicked again. Enter then re-fired that button instead of accepting the staged candidate the strip's own tooltip promised. The exemption is now per key rather than per control: a Button does not consume Left, Right, Delete, F, G, B or 1, so it is no longer handed them. Buttons get Space but deliberately not Enter.
- **The disabled region D buttons were bound to the sampler limitation.** That is null on a backend which supports samplers, so on the default backend the ControlNet button had no tooltip at all — the exact failure this slice exists to prevent — and on the engine it explained samplers on a ControlNet button. They now compose their own reason from what region D has not built plus the backend's own limitation.
- **Switching model or backend silently discarded chosen LoRAs.** The rows' dropdowns are bound to the available-LoRA collection, and clearing it drives each row's selection to null, taking its file path with it. It also removed the "these will not be applied" warning, because by then nothing was configured. A backend that cannot load LoRAs now explains itself without touching the item source, and a genuine refilter re-selects rows by path.
- **Three tests could not fail**, each proven by mutation: the per-model defaults test compared "euler" to "euler" because the fake descriptor fell back to the view model's own initialisers; the LoRA dedup had no coverage the fake could reach; and the Qwen label test compared two nullable lists, which both being null satisfies. All three mutations were re-applied after the fix and each now fails.

The keyboard fix was also verified in the running app: with a panel button focused, G still toggles the grid and does not re-fire the button.

## Verification

- `dotnet build DiffusionNexus.sln -c Debug` — 0 errors.
- `dotnet test DiffusionNexus.Tests` — **5423 passed**, 0 failed, up from 5384 on `develop`.
- **Driven by hand on both backends.** The engine greys the LoRA picker and the sampler and scheduler with its own reasons, keeps the negative prompt live, and adopts Krea 2's 8 steps where the local FLUX.2-klein shows 20. The German-locale guidance label was found this way.

Closes nothing on its own; #518 stays open for regions D, F and G.


DiffusionNexus.UI/Views/DiffusionCanvas/DiffusionCanvasView.axaml
```diff
@@ -7,215 +7,344 @@
...
+
+              <Button Content="Clear canvas" Command="{Binding ClearCanvasCommand}"
+                      ToolTip.Tip="Remove every accepted result from the canvas." />
+
+              <Border Width="1" Margin="6,2" Background="#3A3A3A" />
+
+              <!-- Region D. Disabled, but the tooltip carries the selected backend's own reason rather
+                   than a leaked TODO marker. -->
+              <Button Content="+ ControlNet" Command="{Binding AddControlNetCommand}" IsEnabled="False"
+                      ToolTip.Tip="{Binding SamplerSelectionLimitation}" />
```
Little-God1983: **The `+ ControlNet` tooltip carries the wrong backend reason — and none at all on the default backend.**

The comment above says "the tooltip carries the selected backend's own reason", but the binding is `SamplerSelectionLimitation`, not a ControlNet one.

- On **Diffusion Nexus Core** (the default, `AvailableBackends[0]`) `SamplerSelectionLimitation` is `null` — the local backend supports sampler selection — so this disabled button gets **no tooltip at all**. That is exactly the failure this whole capability mechanism exists to prevent (issue #518: "greying without a reason reads as a bug"), and it is a regression from the old `TODO(v2-controlnet): Add ControlNet conditioning input.` tip.
- On the **Engine** it renders *"The Diffusion Nexus Engine's workflow fixes the sampler and scheduler. Switch to Diffusion Nexus Core to choose them."* on a ControlNet button.

Both backends already author real `BackendFeature.ControlNet` and `BackendFeature.Inpainting` reasons (`StableDiffusionCppBackend.LocalCapabilities` line 71, `ManagedComfyUiBackend.EngineCapabilities` lines 147/149), and `BackendCapabilitiesTests.EveryLimitationTheShippedBackendsDeclareIsAUsableSentence` asserts they are usable sentences — but the view model exposes no `ControlNetLimitation` / `InpaintingLimitation` projection, so those two strings can never reach a user. The `Mask` button below hard-codes region-D text instead, so the inpainting reason is dead too.

Add the two missing projections (plus their `RaiseCapabilityProjections()` entries) and bind each button to its own.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -857,11 +1219,46 @@ private void EndRunEpoch(CancellationTokenSource cts)
...
+    /// reads and what runs would differ.
+    /// </remarks>
+    private BatchSettings CaptureBatchSettings(ModelDescriptor descriptor) => new(
+        Prompt: PromptText,
+        NegativePrompt: string.IsNullOrWhiteSpace(NegativePromptText) ? null : NegativePromptText,
+        Steps: Steps,
+        Cfg: Cfg,
+        Sampler: SelectedSampler,
+        Scheduler: SelectedScheduler,
+        Seed: UseRandomSeed ? null : Seed,
```
Little-God1983: **A locked seed with no value silently generates random, while the panel says random is off.**

`UseRandomSeed = false` with `Seed == null` is a reachable state — `ALockedSeedWithNoValueSaysSoRatherThanShowingNothing` tests it, and the panel has *no way to type a seed*, only "Roll a new fixed seed" and "Reuse last". In that state:

- `SeedText` reads **"Not set"** and the `Random` toggle is **off**;
- this line captures `Seed: null`;
- `RunBatchAsync` line 1303 computes `settings.Seed + i`, and `long? + int` lifts to `null`;
- both backends then roll their own seed (`req.Seed ?? Random.Shared.NextInt64()`).

So the run is random while the UI states it is not — the exact "the panel sends what it shows" invariant this PR is built on, and `LastUsedSeed` then quietly overwrites nothing the user asked for. Either force `UseRandomSeed` back on when `Seed` is null (and say so), or refuse the run with the reason.

Worth pairing with the gap that creates the state: the seed is display-only (`SeedText` in a `Border`), so a seed from a shared generation or an older run cannot be entered at all — only rolled or reused. `ALockedSeedWithNoValueSaysSoRatherThanShowingNothing` locks in the "Not set" wording without covering what Generate then does with it.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -214,24 +396,48 @@ public partial class DiffusionCanvasViewModel : ObservableObject, IDisposable
...
+            Steps = descriptor.DefaultSteps;
+            Cfg = descriptor.DefaultCfg;
+            SelectedSampler = descriptor.DefaultSampler;
+            SelectedScheduler = descriptor.DefaultScheduler;
+        }
 
-    // TODO(v2-loras): show the LoRA picker dialog and add to Loras.
-    [RelayCommand(CanExecute = nameof(AlwaysFalse))]
-    private void AddLora() { /* placeholder */ }
+        _ = LoadLorasForSelectedModelAsync();
```
Little-God1983: **Picked LoRA rows survive a model change and are then applied to the new model.**

`OnSelectedModelChanged` refills `AvailableLoras` for the new model's base-model labels, but never touches `Loras` — the rows the user already picked. Repro:

1. Model = FLUX.2-klein, add a LoRA row, pick a `Flux.2 Klein 9B` LoRA.
2. Switch the model dropdown to `qwen-image-2512`.
3. `AvailableLoras` is refiltered to `Qwen`/`Qwen 2`, but the Flux row stays in `Loras` with `IsEnabled = true` and its `FilePath` intact.
4. Generate → `ResolveLoraReferences` (line 838) emits it, `CaptureBatchSettings` puts it in `DiffusionRequest.Loras`, and the local backend loads a Flux LoRA over Qwen.

The only mismatch that is warned about today is the *backend* one (`settings.Loras.Count > 0 && !IsLoraSupported`, line 941). The model mismatch — which is the one the whole `ModelBaseModelLabels` map exists to prevent — passes silently.

Related: `AvailableLora` is a record whose `Thumbnail` is a `Bitmap`, so the refilled list contains value-unequal instances; the picker's `AutoCompleteBox` may also null `SelectedItem`, which `OnSelectedLoraChanged` turns into `FilePath = null` — a row that looks picked but resolves to nothing. Either clear/disable incompatible rows on model change, or filter them out in `ResolveLoraReferences` with a warning.

DiffusionNexus.UI/Views/DiffusionCanvas/DiffusionCanvasView.axaml.cs
```diff
@@ -170,9 +170,24 @@ private void OnPreviewKeyUp(object? sender, KeyEventArgs e)
...
+    /// </remarks>
+    private bool ShouldLeaveTheKeyAlone()
+    {
+        var focused = TopLevel.GetTopLevel(this)?.FocusManager?.GetFocusedElement();
+        if (focused is TextBox)
+            return true;
+
+        return focused is Visual visual
+            && GeneratePanel is { } panel
+            && (ReferenceEquals(visual, panel) || visual.GetVisualAncestors().Contains(panel));
```
Little-God1983: **The new guard is too broad: it kills *every* canvas shortcut whenever focus is anywhere in the generate panel — including Enter/Delete/←/→ for the staging strip, immediately after Generate is pressed.**

The panel needed exempting for sliders, combos and the spinner, but this exempts *all* keys for *any* focus inside `GeneratePanel`. Clicking the `Generate` button focuses it (Avalonia focuses focusable controls on pointer press), and that button is inside `GeneratePanel`. So the moment the batch finishes and candidates appear in the strip, the user's focus is still on Generate and:

- `Enter` (accept), `Delete` (discard), `←`/`→` (previous/next) — the entire staging workflow — are dead;
- so are `F`, `1`, `B`, `G`, `Escape` and `Space`/compare.

`Shortcuts.md` now documents this as "Click the canvas to get the shortcuts back", but making the primary post-Generate keyboard flow require a mouse click is a real regression, not a documented limitation.

Narrow it to the keys the focused control actually consumes — e.g. exempt only when `focused is TextBox or Slider or ComboBox or NumericUpDown or ToggleButton or CheckBox` (still an ancestor check for templated parts), or return focus to `Surface` after `GenerateCommand` completes.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -505,6 +739,124 @@ public async Task LoadModelsAsync()
...
+
+        if (labels.Count == 0)
+        {
+            Publish([], $"{SelectedModel?.DisplayName ?? modelKey} has no published LoRA base model, so none can be matched.");
+            return;
+        }
+
+        try
+        {
+            var loras = await _loraCatalog.GetInstalledLorasAsync(labels).ConfigureAwait(false);
```
Little-God1983: **The LoRA reload leaks decoded thumbnails and runs 2–3 times per selection change, with no cancellation.**

`LoraCatalog.GetInstalledLorasAsync` constructs a **fresh** `Bitmap` per LoRA on every call (`LoadThumbnailAsync` → `new Bitmap(ms)`); there is no cache. This view model is a **DI singleton** (`App.axaml.cs` line 720) and now re-runs that query on every model change *and* every backend change:

- `AvailableLoras.Clear()` (line 763) drops the previous `Bitmap`s **without disposing them** — the file disposes `Frames` bitmaps explicitly everywhere else, and `Dispose()` never touches `AvailableLoras` either;
- a load superseded by the generation counter never even reaches the collection, so its whole decoded set is discarded on the floor;
- switching to the Engine fires this twice (once via `SelectedModel = …` inside `OnSelectedBackendChanged` → `OnSelectedModelChanged`, then again at line 414); the Local path additionally fires it once with the *stale* engine model key before `LoadModelsAsync` lands, and `AvailableModels.Clear()` nulls `SelectedModel` for a third pass.

Also, the `CancellationToken` overload is available and unused — a superseded load still walks the model DB and decodes every BLOB before its result is thrown away.

Pass a per-generation token, and dispose the outgoing thumbnails (or have the catalog cache them) before `Clear()`.

DiffusionNexus.UI/Services/Lora/ModelBaseModelLabels.cs
```diff
@@ -0,0 +1,92 @@
...
+        // Krea 2 deliberately maps to nothing. Civitai has no Krea 2 base-model label — its "Flux.1 Krea"
+        // entry is Flux.1 Krea dev, an unrelated model — so there is no correct value to filter on. An
+        // empty list means "no compatible LoRAs can be identified", which the caller must surface as an
+        // explanation. It must NOT be turned into a null filter: ILoraCatalog reads null as "return
+        // everything", which on a real library is thousands of rows each decoding a thumbnail.
+        [KreaModelKey] = [],
+    };
+
+    /// <summary>The engine backend's Krea 2 model key. Duplicated as a literal to avoid a UI-layer cycle.</summary>
+    private const string KreaModelKey = "krea2";
```
Little-God1983: **The duplicated `"krea2"` literal is justified by a cycle that does not exist.**

> `/// <summary>The engine backend's Krea 2 model key. Duplicated as a literal to avoid a UI-layer cycle.</summary>`

`ModelBaseModelLabels` is `DiffusionNexus.UI/Services/Lora/` and `ManagedComfyUiBackend` is `DiffusionNexus.UI/Services/Diffusion/` — **the same assembly**. There is no project cycle and no namespace cycle (namespaces do not create dependencies); `DiffusionCanvasViewModel` already `using`s both. `ManagedComfyUiBackend.Krea2Model.Key` is directly reachable.

As written, renaming the engine's model key silently turns this entry into a dead map row: `ForModelKey("krea2-turbo")` would return `null`, the canvas would show *"No LoRA compatibility is recorded for … yet"* instead of the authored "Civitai has no Krea 2 label" explanation, and nothing would fail to compile. Reference the constant.

DiffusionNexus.UI/Services/Diffusion/ManagedComfyUiBackend.cs
```diff
@@ -185,6 +212,28 @@ public async IAsyncEnumerable<DiffusionStreamItem> GenerateAsync(
         ArgumentNullException.ThrowIfNull(request);
         cancellationToken.ThrowIfCancellationRequested();
 
+        // This backend submits ONE graph — the Krea 2 workflow — regardless of what is asked of it, and
+        // it never reads the descriptor. Meanwhile its catalog unions Krea 2 with a real disk scan of the
+        // engine's models root (EngineModelCatalog), so after one readiness probe the caller's model
+        // dropdown can legitimately offer Z-Image-Turbo, FLUX.2-klein or Qwen. Without this guard,
+        // picking one of those ran the Krea 2 GGUF and reported the result under the other model's name:
+        // a wrong image, labelled convincingly. Refused as data, matching this seam's error contract.
+        if (!string.Equals(request.ModelKey, Krea2Model.Key, StringComparison.Ordinal))
```
Little-God1983: **The refusal is the right last line of defence, but the list that produces the bad selection is still advertising those models.**

`EngineModelCatalog.ListAvailable()` (line 82) unions Krea 2 with the disk scan, and `OnSelectedBackendChanged` fills the canvas dropdown straight from it. So on the Engine backend the model combo *still* offers `z-image-turbo`, `flux2-klein` and both Qwen models. The user picks one, the panel adopts that descriptor's sampling defaults (`OnSelectedModelChanged`), they tune the prompt and press Generate — and only then land here.

The deeper fix is one level up: `ListAvailable()` should return only what this backend can actually submit (i.e. `Krea2Model`), while `TryGet` keeps resolving discovered keys for any caller that genuinely needs a descriptor. Otherwise every future consumer of `Catalog.ListAvailable()` re-inherits the same trap, and this guard has to be re-added at each new call site.

Minor, in the message itself: `'{request.ModelKey}' runs on Diffusion Nexus Core` is asserted for *any* non-`krea2` key, including one Core cannot run either.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -214,24 +396,48 @@ public partial class DiffusionCanvasViewModel : ObservableObject, IDisposable
...
+    /// its own defaults; the panel now has to reproduce that faithfully.
+    /// </remarks>
+    partial void OnSelectedModelChanged(CanvasModelOption? value)
+    {
+        if (value?.Descriptor is { } descriptor)
+        {
+            Steps = descriptor.DefaultSteps;
+            Cfg = descriptor.DefaultCfg;
+            SelectedSampler = descriptor.DefaultSampler;
+            SelectedScheduler = descriptor.DefaultScheduler;
```
Little-God1983: **The model's default sampler/scheduler is adopted without checking it exists in the offered list — the inverse of the hole `SupportedSamplers` was added to close.**

`AvailableSamplers` is deliberately restricted to `MapSampler`'s own vocabulary so a user cannot pick a dead sampler. But this assigns `descriptor.DefaultSampler` / `DefaultScheduler` straight into the bound property with no membership check. `ModelDescriptor`'s own doc gives `"dpm++2m_karras"` as an example default, and `MapSampler` has no such case.

For any descriptor whose default is outside the list, the `ComboBox` finds no matching item, sets `SelectedItem` to `null`, and the TwoWay binding writes `null` back into the non-nullable `SelectedSampler`. Result: the sampler dropdown goes blank, `SamplingSummary` renders `" · 20 · cfg 7.0"` with a leading gap, and `CaptureBatchSettings` sends `Sampler: null` — the model's real default silently comes back from the backend fallback while the header shows nothing.

Every shipped descriptor happens to use `euler`/`simple` today, so this is latent; guard it (`AvailableSamplers.Contains(...) ? … : "euler"`, or log and keep the current value) rather than relying on that.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -505,6 +739,124 @@ public async Task LoadModelsAsync()
...
+        }
+
+        var resolved = new List<LoraReference>();
+        foreach (var row in Loras)
+        {
+            if (!row.IsEnabled || string.IsNullOrWhiteSpace(row.FilePath))
+                continue;
+            if (!seen.Add(row.FilePath))
+            {
+                EmitInfo($"Skipping '{row.DisplayName}' — the model already applies it.");
```
Little-God1983: **This message is wrong for the case the tests actually cover.**

`seen` is pre-seeded with `descriptor.DefaultLoras`, but it is also the running dedup set for the user's own rows. So the *second* of two rows pointing at the same file — precisely what `TheSameLoraPickedTwiceIsAppliedOnce` exercises — is reported to the Unified Console as *"the model already applies it"*, which is false: the model applies nothing, the user picked it twice.

Track the two sets separately (a `defaults` set checked first, then a `seen` set) so the log can say "the model already applies it" vs "already added above". Otherwise the console misdirects anyone debugging why their second row did nothing.

DiffusionNexus.Inference/StableDiffusionCpp/StableDiffusionCppBackend.cs
```diff
@@ -274,9 +298,16 @@ private static GenerationOutcome RunGeneration(SDNet.DiffusionModel model, Model
...
         }
 
-        // TODO(v2-negative-prompt): apply req.NegativePrompt via .WithNegativePrompt(...) once enabled.
-        // TODO(v2-controlnet):      apply req.ControlNets via .WithControlNet(image, strength).
-        // TODO(v2-inpaint):         apply req.MaskImage via .WithMaskImage(...) for inpaint flows.
+        // A blank negative prompt is left off entirely rather than sent as "": the native default and an
+        // empty string are not guaranteed to be the same conditioning, and "the user typed nothing"
+        // should mean "unchanged", not "an empty negative".
+        if (!string.IsNullOrWhiteSpace(req.NegativePrompt))
+            genParams = genParams.WithNegativePrompt(req.NegativePrompt);
```
Little-God1983: **Stale contract docs left behind by this fix.**

Wiring `WithNegativePrompt` here (and node 35 on the engine side) makes `DiffusionRequest`'s own documentation false — `DiffusionRequest.cs` still carries:

```csharp
// TODO(v2-negative-prompt): Backend currently ignores this field — placeholder for the negative-prompt UI control.
/// <summary>Negative prompt. <b>Currently ignored by the v1 backend.</b></summary>
public string? NegativePrompt { get; init; }
```

That is now the single most authoritative place a future backend author will look, and it tells them the field is dead. Since this PR's whole thesis is that a capability must be stated where it is knowable, the request contract should say "honoured by both backends" and the TODO should go. (`Loras` has the same problem in reverse — its doc says "applied per-generation by the stable-diffusion.cpp backend" with no hint the engine drops it, which is now knowable from `EngineCapabilities`.)

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -150,36 +173,195 @@ public partial class DiffusionCanvasViewModel : ObservableObject, IDisposable
     [ObservableProperty]
     private bool _isRegionOccupied;
 
-    #region v2 Placeholder properties (bound to disabled UI controls)
+    /// <summary>
+    /// The two-word form of <see cref="RegionModeText"/> for the panel's header chip, which has about
+    /// 150px. The long form is the chip's tooltip, so nothing is lost.
+    /// </summary>
+    [ObservableProperty]
+    private string _regionModeBadge = "Text to image";
```
Little-God1983: **`RegionModeBadge` is derivable state stored twice.**

It is written only at line 608, from the same `overlapping > 0` expression that already sets `IsRegionOccupied` one line above, and read only by the header chip. A computed property removes the field, the initialiser (which can disagree with the real box state until `RefreshRegionMode()` first runs), and the extra `PropertyChanged`:

```csharp
public string RegionModeBadge => IsRegionOccupied ? "Image to image" : "Text to image";
```

with `[NotifyPropertyChangedFor(nameof(RegionModeBadge))]` on `_isRegionOccupied` — the pattern the rest of this file uses for `SamplingSummary`, `GuidanceText` and `DenoiseText`. (Same nit one level down: `PromptLengthText` is raised by a hand-written `OnPromptTextChanged` partial at line 281 where every other derived string uses the attribute.)

DiffusionNexus.UI/Views/Controls/DiffusionCanvasSurface.cs
```diff
@@ -343,7 +350,18 @@ private void OnRastersCollectionChanged(object? sender, NotifyCollectionChangedE
...
     private void OnRasterPropertyChanged(object? sender, PropertyChangedEventArgs e) => InvalidateVisual();
 
-    private void OnObservedBoxChanged(object? sender, EventArgs e) => _boxLayer.InvalidateVisual();
+    private void OnObservedBoxChanged(object? sender, EventArgs e)
+    {
+        _boxLayer.InvalidateVisual();
+
+        // Selecting a model re-snaps the box onto that model's lattice, and the grid draws the same
+        // lattice, so it has to be re-recorded when that value moves.
+        if (Box is { } box && box.Alignment != _gridAlignment)
```
Little-God1983: **The new `_gridAlignment` tracker is never resynced when the `Box` property itself changes.**

`OnPropertyChanged`'s `BoxProperty` branch (line 276) calls `AttachBox` and invalidates only `_boxLayer` — it does not touch `_gridAlignment` or `InvalidateVisual()`. So a surface that attaches a box already sitting on a 16 lattice starts with `_gridAlignment == 64`: the canvas view model is a DI singleton, so its `Box` survives navigation with `Alignment` already set by the last Generate, and every re-created surface re-enters that state.

The visible grid is still correct (`DrawGrid` reads `Box?.Alignment` live), but the tracker is stale, so the *first* unrelated `Changed` — one pointer move of a box drag — sees `16 != 64` and fires a **full-surface** repaint, which is exactly what the `_boxLayer` split and this tracker exist to avoid. Set `_gridAlignment` from the new box inside `AttachBox`.

DiffusionNexus.Inference/Abstractions/BackendCapabilities.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <param name="limitations">
+    /// One entry per <b>unsupported</b> feature, mapping it to the one-line reason shown at the control.
+    /// Anything absent from this map is supported. Stating it negatively keeps a newly added
+    /// <see cref="BackendFeature"/> supported-by-default rather than silently disabling it everywhere,
+    /// which would be the wrong failure direction: a feature that works but is greyed out is invisible,
+    /// while one that is offered and does nothing is at least reportable.
+    /// </param>
+    public BackendCapabilities(IReadOnlyDictionary<BackendFeature, string>? limitations = null)
+    {
+        _limitations = limitations ?? new Dictionary<BackendFeature, string>();
```
Little-God1983: **The capability set keeps the caller's dictionary instead of copying it.**

`_limitations` aliases whatever `IReadOnlyDictionary` was passed in. `IReadOnlyDictionary` is not immutable — a `Dictionary<BackendFeature, string>` upcast to it stays mutable through the original reference.

Both shipped sets are `static` singletons handed out through `LocalCapabilities` / `EngineCapabilities` and reused by every `CanvasBackendOption`, so anyone who holds the builder dictionary can silently re-gate every canvas in the process. A `.ToDictionary(...)` (or a `FrozenDictionary`) in the constructor makes the type actually immutable, which is what the whole "anything absent is supported" contract assumes.

DiffusionNexus.UI/ViewModels/DiffusionCanvas/DiffusionCanvasViewModel.cs
```diff
@@ -150,36 +173,195 @@ public partial class DiffusionCanvasViewModel : ObservableObject, IDisposable
...
+    /// <summary>The sampling algorithm. Only meaningful on a backend that honours it.</summary>
     [ObservableProperty]
+    [NotifyPropertyChangedFor(nameof(SamplingSummary))]
     private string _selectedSampler = "euler";
 
-    // TODO(v2-loras): observable list bound to the LoRA picker (each item carries path + strength).
-    public ObservableCollection<object> Loras { get; } = [];
+    /// <summary>The noise schedule. Same backend caveat as <see cref="SelectedSampler"/>.</summary>
+    [ObservableProperty]
+    private string _selectedScheduler = "simple";
```
Little-God1983: **The scheduler is invisible in the collapsed state, and its changes do not refresh the header.**

The sampling section is collapsed by default and its header is documented as the only place the values stay readable — but `SamplingSummary` is `"{sampler} · {steps} · cfg {cfg}"`, with no scheduler. `_selectedScheduler` is also the one panel value with no `[NotifyPropertyChangedFor]` at all.

Since the scheduler *is* honoured by the local backend and *is* sent in the request, a user who changed it has no way to see that from the collapsed panel — the same argument that put steps and cfg in the summary. Either add it (`euler · karras · 9 · cfg 1.0`) or drop the control.

Into-The-Latent:
## Recheck of the 14 inline findings against `dc5c6099`

The head moved after I posted: `dc5c6099` ("fix(canvas): address the review of the generate panel")
landed and closes **three** of the fourteen. I re-verified all fourteen against the code at that
commit rather than at `965c246e`, and found two things the fix did not cover — one of which it
made visible.

### Fixed — please disregard those three comments

| Comment | What closed it |
|---|---|
| `+ ControlNet` tooltip bound to `SamplerSelectionLimitation` | `ControlNetTooltip` / `MaskTooltip` now compose a region-D sentence with the backend's own `ControlNet`/`Inpainting` limitation, and both are raised in `RaiseCapabilityProjections()`. |
| Picked LoRA rows survive a model change | `RepublishAvailableLoras` re-selects surviving rows by path and reports the rest; the `!IsLoraSupported` path now uses `Explain` so a backend switch does not empty the item source at all. Moving the capability check ahead of the catalog check is right. |
| Tunnel guard exempts every key inside the panel | The guard is per key now. The reasoning in the new remarks is better than my finding was: focus *stays* on a clicked Button, so the broad version broke the flow on any panel click, not only after Generate. Withholding `Enter` from `Button` while keeping `Space` is the correct call. |

### Still stands, exactly as posted (7)

- **`ManagedComfyUiBackend` refusal sits below the list that produces the bad selection.**
  `EngineModelCatalog.ListAvailable()` still unions Krea 2 with the disk scan, and
  `DiffusionCanvasViewModel.cs:428` fills the engine dropdown straight from it. The user still
  picks FLUX.2-klein, still gets its 20-step defaults adopted, and still only learns at Generate.
- **LoRA thumbnails leak and re-decode.** `LoraCatalog.LoadThumbnailAsync` does `new Bitmap(ms)`
  per row with no cache; `RepublishAvailableLoras` still calls `AvailableLoras.Clear()` with no
  disposal and `Dispose()` still does not touch the collection, on a DI-singleton view model
  (`App.axaml.cs:720`). `GetInstalledLorasAsync` is still called without a token, so a superseded
  load decodes a full set that `Publish` then discards on the generation check.
- **The `"krea2"` literal's stated justification is false** — `ManagedComfyUiBackend` and
  `ModelBaseModelLabels` are both in `DiffusionNexus.UI`, so `Krea2Model.Key` is directly
  reachable and there is no cycle. See the sharpening below.
- **`seen` doubles as the defaults set and the dedup set**, so a user's own duplicate row is still
  logged as "the model already applies it" — false, and it points a reader at `DefaultLoras`.
- **`DiffusionRequest` still documents the opposite of what now ships**: line 53's
  `TODO(v2-negative-prompt)` and line 54's "**Currently ignored by the v1 backend.**" describe a
  field both backends now honour (`StableDiffusionCppBackend.cs:305`,
  `Krea2WorkflowPatcher.cs:104`), and the type summary still lists `NegativePrompt` among the
  fields "wider than v1 honors". `Loras` has the mirror gap.
- **`_gridAlignment` is never resynced in `AttachBox`.** `Render` reads `Box?.Alignment` directly
  so the drawn grid is always right — the cost is one spurious full-surface `InvalidateVisual` on
  the first `Changed` after a re-created surface picks up the singleton's already-16-aligned box,
  which is what the `_boxLayer` split exists to avoid.
- **The scheduler is absent from `SamplingSummary`** while the Expander is `IsExpanded="False"`,
  so a value the local backend honours and `CaptureBatchSettings` sends has no collapsed reading.

### Still stands, but I over-stated the harm (3)

- **Seed.** `UseRandomSeed = false` with `Seed == null` does still send null and both backends roll
  their own (`ManagedComfyUiBackend.cs:259`, `StableDiffusionCppBackend.cs:237`). But I wrote that
  the panel "states it is locked" — it does not: `SeedText` correctly reads **"Not set"**. The real
  defect is narrower: the Random toggle's own tooltip promises "Turn this off to keep one and vary
  the prompt instead", and turning it off without rolling delivers no such thing, silently. A
  warning, or having the toggle roll a seed on its way off, would be enough.
- **Default sampler not checked against `AvailableSamplers`.** Latent as posted, but the
  consequence is smaller than I said: a null `Sampler` reaching the request hits
  `req.Sampler ?? d.DefaultSampler` in the local backend, so the *right* sampler still runs. The
  harm is a blank dropdown and a `SamplingSummary` reading " · 20 · cfg 7.0".
- **`BackendCapabilities` aliases the caller's dictionary.** True, and worth one `ToDictionary`,
  but not exploitable today: both `LocalCapabilities` and `EngineCapabilities` pass an inline
  dictionary literal, so no caller retains the reference. Hardening, not a bug.

`RegionModeBadge` / `PromptLengthText` stand as posted and were only ever simplification nits.

### Two the fix did not cover

**1. The ControlNet and Inpainting escape hatches point at a backend that cannot do it either —
and `dc5c6099` is what made the text visible.**

On Diffusion Nexus Core the tooltip now composes to:

> Control layers arrive with the layer stack (issue #518 region D). Diffusion Nexus Core has no
> ControlNet path for these models. **Switch to the Diffusion Nexus Engine to use control layers.**

but `EngineCapabilities` declares `ControlNet` unsupported too ("Control layers are not wired into
the engine's workflow yet"). Inpainting is the same pair. `BackendCapabilities`' own doc sets the
rule this breaks — limitation text "should name the backend that would honour the feature, so the
sentence carries the way out". Here it names one that does not, which is the same class of
mislead the capability surface exists to prevent, just one level up. Before `dc5c6099` these two
sentences never reached a user; the tooltip fix routed them to the screen. Either drop the "switch
to" clause from both `LocalCapabilities` entries, or gate composing it on the other backend
actually supporting the feature.

**2. `RepublishAvailableLoras` leaves the actual clearing to the ComboBox.**

For a row whose LoRA does not survive the new filter it increments `dropped` and warns that the
row "**was cleared**" — but never sets `Loras[i].SelectedLora = null`. The clearing is a
side-effect of `AvailableLoras.Clear()` writing back through the ComboBox's TwoWay `SelectedItem`,
so it happens only for rows whose ComboBox is realized. `FilePath` is derived from `SelectedLora`
(`LoraPickerItemViewModel.OnSelectedLoraChanged`), so any row that misses the write-back keeps its
path and `ResolveLoraReferences` still sends it — the exact incompatible-LoRA case this method was
added to prevent. An explicit `Loras[i].SelectedLora = null;` in the `else` branch closes it and
makes the invariant assertable without a view.

**3. (sharpens the `"krea2"` finding.)** The new "every model the canvas can select has an entry"
test hardcodes `"krea2"` as a literal as well, so the string now exists in three places. Renaming
`Krea2Model.Key` still leaves the map row dead, `ForModelKey` returning null, and the test green —
the guard does not close the hole its own commit message credits it with.

---

Nothing here blocks; the three fixes are good and the reasoning in `dc5c6099` is sound where it
corrects mine. Item 1 above is the one I would not ship as-is, since it is a user-facing false
statement in the feature whose whole point is not making them.

